PR #1, opened 2016-06-04
Use direct link to Newsletter

LukasReschke:
Link directly to the subscribe form. As brought up by @jancborchardt 😉 

cc @jancborchardt @jospoortvliet

### DIFF
--- a/page-homepage.php
+++ b/page-homepage.php
@@ -50,7 +50,7 @@
 	</div>
 	<div class="row pad-bottom">
 		<div class="col-md-8 col-md-offset-2  text-center">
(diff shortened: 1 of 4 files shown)


mar1u5:
👍 Makes sense, since users who want to unsubscribe can fetch the link to unsubscribe from the newsletters 😉 ...what about directly embed the subscription into the website?

jospoortvliet:
:+1: awesome indeed...  And @mar1u5 is right, it'd be even nicer to be able to subscribe right there and then but I have no clue how to make that happen so until then, this'll do the trick...

mar1u5:
http://docs.phplist.com/CustomSubscribeForm.html @jospoortvliet 😉

jospoortvliet:
I will put that in the issue list then...

jancborchardt:
Awesome – yes great point by @mar1u5 to directly integrate it :)

PR #7, opened 2016-06-09
Do not add the X-Mailer header

LukasReschke:
Our SpamAssasin configuration gives that a penalty of 2.5 points, 3 are required to be considered spam. Thus a lot of mails coming from here are marked as spam…

> 2.5 PHP_ORIG_SCRIPT        Sent by bot & other signs

@karlitschek @jospoortvliet This should fix the spam issue ;-)


karlitschek:
👍

PR #10, opened 2016-06-09
Remove config.php from repo

vincchan:
Finish off config.php based on 2e4e044778dabc7f31f8517fdf56fed733dfaba7

@jospoortvliet what do you think? Will this be a problem for existing instances as it will delete those config.php too

### DIFF
--- a/config.php
+++ b/config.php
@@ -1,7 +1,0 @@
-<?php
-// Enables CONTRIBOOK (required mysql database)
-define(CONTRIBOOK, 0);
-
-// Defines the author id of the official user posting annoucenments (used to produce rss feed)
-define(OFFICIALAUTHORID, 2);
-?>


mar1u5:
Makes sense, since the 5. step of the setup-instructions for test environments is:

> Copy over config.php.sample to config.php and adjust settings as necessary (defaults will work just fine for local environments)

However ... If we remove **config.php** completely, we could also remove the line from **.gitignore**

cc @jospoortvliet

vincchan:
@mar1u5 No, .gitignore is fine because if there was is not that entry, it will simply be committed again.

mikebabb:
Agreed; if it's deleted and removed from gitignore as well it'll end up back in the repo as a new file when a new developer begins work on the site. I forgot to include file deletion in the commit that added the gitignore entry.

mar1u5:
Done :wink:

PR #11, opened 2016-06-09
Swaped to the same background image as server

### LINKED ISSUE #9: Make the front page prettier
So Jan had a great idea... Christoph Wurst did some nice design work using a photo Frank made and we have this new look added in this pr: https://github.com/nextcloud/server/pull/28

Nice Nextcloud blue right? Indeed, we can use this for our front page. This isn't terribly hard, an image to replace and a bit of CSS change on the button perhaps to make it a white outline. @jancborchardt also suggested to semi-bold the font.

vincchan:
fixes #9


mar1u5:
:+1: ...THX!

jospoortvliet:
Oh, it looks awesome:
screenshot_20160610_085630

For the button I'd say we should make the center transparent and a thicker white outline. Something like this:
screenshot_20160610_090324

I made the outline white, 5px and the font-family Open Sans Bold. Also, the background of the button transparent. What'd yah think?

jospoortvliet:
@jancborchardt ;-)

vincchan:
@jospoortvliet alright, do you wanna take over or shall I alter the button?

mar1u5:
Would be nice if you can 😉 @vincchan

vincchan:
@mar1u5 @jospoortvliet here you go. I made the border 3px as I found 5px a bit too much. Any objections?

jospoortvliet:
It looks great, just tested. I think that making the font bold or semi-bold would be nice, too...

jospoortvliet:
.. but if you don't like it, that's fine. Then we merge as-is and let @jancborchardt tell us what he thinks later :dancers:

PR #12, opened 2016-06-10
fix styles & fonts

jancborchardt:
Fix the font stack to use humanist (Open Sans, Frutiger) rather than realist (Helvetica) font style because Open Sans is also humanist. We use the same font stack in Nextcloud server.

Also switching to Open Sans Semibold instead of Bold. @jospoortvliet I already added it to the repo.

And lastly fixing the position of the background image to center. Follow-up to @vincchan’s https://github.com/nextcloud/nextcloud.com/pull/11

Please review @nextcloud/designers :)

### DIFF
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -27,7 +27,7 @@
 	font-family: 'Open Sans';
 	font-style: normal;
 	font-weight: bold;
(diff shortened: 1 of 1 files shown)


vincchan:
@jancborchardt looks good to me 👍

jancborchardt:
@jospoortvliet do you want to deploy this? Not really a reason to wait for this til the release, right?

PR #23, opened 2016-06-13
some updates

jospoortvliet:
would be nice to have some design fu on this one ;-)

### DIFF
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -547,8 +547,16 @@ p.federation-id {
     
 }
 
+/*---- Enterprise page ---- */
 
+.table-center {
+	text-align: center !important;
+}
 
+.enterprise-icon {
+	font-size:80px;
+	color: #0082c9;
+}
 
 /*---- history page ---- */
 
--- a/page-enterprise.php
+++ b/page-enterprise.php
@@ -2,46 +2,46 @@
 <div class="page-header">
 	<h1>Privacy and control for your organization</h1>
 </div>
-<h2>Subscription options</h2>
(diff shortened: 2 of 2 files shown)


jospoortvliet:
What it looks like now:

enterprise - nextcloud

@jancborchardt

jospoortvliet:
<Blizzz> add 1-50 users, community support, community consulting, no commercial support, FREE

jancborchardt:
Mostly the table could use a bit more info what differentiates the tiers. Currently it’s really only the support range. Is there nothing else? (Having the deployment and security consulting be the same doesn’t help there. ;)

Do or should we offer anything else specifically? cc @karlitschek

jospoortvliet:
Yeah, there simply is nothing else at the moment. There are no tiers, actually, it's just that if you have more users the cost/user goes down ;-)

blizzz:
> Yeah, there simply is nothing else at the moment. There are no tiers, actually, it's just that if you have more users the cost/user goes down ;-)

You could merge the cells and write it down only once, or have a description on the left side and work with green checkmarks and red crosses, or something like that. However, there are really few things for a comparison, making me doubtful whether it really has value or keeps people off. Maybe the wall of text can be reduced and aspects soaked in here? Another think to consider: with the additional FREE column, the central ones can be merged and discount ranges mentioned in the table.

MorrisJobke:
And maybe change `Eur 3200` to `3200 €`

jospoortvliet:
So we still have to figure out how/what/where to sell, let's do that before we create a webpage ;-)

Will close this PR for now.

PR #25, opened 2016-06-14
install page preparation

jospoortvliet:
Current look. Ideas for improvement?
install - nextcloud


jospoortvliet:
screenshot_20160614_111802

jancborchardt:
Instead of the small icons next to the headings, let’s put the icons centered above the title. And then use proper h2 for the headings so they are bigger and not bold. Basically just like the 3 boxes on the front page.

Also, instead of putting the numbers in the blue circles (which is very unreadable) I would basically switch numbers and icons – putting the numbers before the titles.

jancborchardt:
And regarding the installation popup – maybe something for later, but the mess of information is really crazy here.

- veryfying via MD5 and PGP is just optional, not necessary steps
- all the info has the same font-size, making it confusing

The important info should be pulled out of the popup, and all the less important stuff can be kept in there:
- have the »Download« button on the main install page actually directly download the package (just use either tar.bz2 or zip, I don’t care which but choose _only one_)
- have a »Documentation« button next to that download button
- have a »Details…« text below the buttons which opens the popup with all that extra window

jospoortvliet:
install - nextcloud 3
install - nextcloud 2
install - nextcloud 1

jospoortvliet:
install - nextcloud 4

jospoortvliet:
Unless there are serious issues or very easy fixes, I'd like to get this live so I can focus on the enterprise page ;-)

jancborchardt:
I wouldn’t force a 4-column layout and press it all in one row. It was fine with 3 on top and 2 below as before.

Or maybe even simpler: 2 big on top (1. Download and 2. Sync). And then the help boxes below that with more whitespace.

jospoortvliet:
So I made it a 3-2, looking good now. Merging!

PR #44, opened 2016-06-30
first providers

jospoortvliet:
these are already available. TODO:

* [x] add link on download page


jospoortvliet:
What it looks like right now:

providers - nextcloud

mar1u5:
Does the **WK IT** - Logo meet our policy because of its size :confused:

jospoortvliet:
@mar1u5 honestly no, I just resized it and fit it in the size we want (600x150). I could've asked them to get me a new one but this was faster ;-)

mar1u5:
Is it a good idea to show the texts per default for every Provider. The ownCloud Provider page does not too :grin: ...what about a popup when you click on the logo :wink:

jospoortvliet:
@mar1u5 the problem with the old providers page was that it had no way of filtering the huge list of providers. Now you can filter easily so I thought it is nicer to give some more differentiating information - it used to simply be a dumb list of images. Now there's some info they can add. And yeah, if they make their logo's ugly, Not My Problem.

jancborchardt:
Yup, looks cool as first iteration! Just one thing: Can we reduce the whitespace below the description a bit? That looks strange.

jospoortvliet:
yeah, thing is, it has to cover a 'worst case' of the max nr of characters, or things start to look real weird. I could remove the outline or shorten-for-now and see how things go... Let me do the latter, that is quick and easy ;-)

jancborchardt:
Yeah, removing the outline makes it look better anyway. Less lines, just use whitespace :)

PR #47, opened 2016-07-01
make YT look better!

jospoortvliet:
This makes YT videos look a LOT better - basically just like 'real' embedded youtube videos, but without the big performance hit they cause page loads :dancer: 

/me is quite proud of this one even though he spent more time on it than he should have
Old/new:

spreedbox improvement

(in the old one, hovering the video wouldn't make the icon go red, only hovering the actual icon. Now, hovering the area already works - as it should)

### DIFF
--- a/assets/css/styles.css
+++ b/assets/css/styles.css
@@ -1097,40 +1097,61 @@ img.desaturate {
 	
 }
 
-/* image overlay */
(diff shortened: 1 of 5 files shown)


jancborchardt:
Looks awesome! :+1:  :)

jospoortvliet:
Thanks :D

PR #48, opened 2016-07-01
This is Nextcloud! - Provider apply 

mar1u5:
Could you please check the following @jospoortvliet :

- [ ] Check whether Nextcloud Inc. as company is correct (line 63)  - it was GmbH, wasn't it 😜
- [ ]  Do we still need **page-providersubmit.php** and **apply2.php** 😕

### DIFF
--- a/page-apply.php
+++ b/page-apply.php
@@ -4,7 +4,7 @@
 </div>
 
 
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Good catch. And yes, we need providersubmit so providers can apply and apply2 will be for consultants/partners, coming soon...

PR #56, opened 2016-07-02
Put RC on our install page

jospoortvliet:
spectacle j17468


jospoortvliet:
Ok, Frank likes it, what more do you need... going live!

PR #57, opened 2016-07-04
Features page, first draft!

jospoortvliet:
Quite a start on the features page... All kinds of feedback needed, PR's super welcome with that ;-)
features - nextcloud


jancborchardt:
Looking good! :+1: we should simplify it a bit to make it easier to read (and probably add icons to make it quicker to understand) – but that's future stuff!

MorrisJobke:
I really like this 👍

PR #58, opened 2016-07-05
Enterprise

jospoortvliet:
Update to the enterprise page with pricing and all that.

This also updates font-awesome and bootstrap.


LukasReschke:
👍

PR #61, opened 2016-07-06
fsck with jquery and css to make yt videos a bit more viewport aware

jospoortvliet:
it isn't perfect yet but the video's look a lot better on small devices and aren't broken on a normal desktop...


assets/css/styles.css
```diff
@@ -1298,3 +1298,22 @@ img.desaturate {
 /*	width: 100%;
 	height: 100%;*/
 }
+
+.YouTubeDialog {
+	width: 50% !important;
+	height: 50% !important;
+	padding: 5px;
+/* 	padding-top: 15% !important; */
```
blizzz: necessary to keep the comment? then doc would be cool.

jospoortvliet:
removed. Going once.... Twice... Gone!

Issue #46, opened 2016-06-30
Switch earlier to the responsive "sandwich" menu 


mar1u5:
With Safari on my iPad Pro 9.7, the menu is in a wrong position, if i take the iPad horicontal. I think we should go for the "sandwich" menu here :wink: 

![image](https://cloud.githubusercontent.com/assets/19711361/16495521/2b11644a-3eee-11e6-8788-ac268f5e0f39.png)

@nextcloud/designers

jospoortvliet:
probably a tiny sizing issue but - no idea how to fix it without being able to use 'inspect' on an ipad :-1:

mikebabb:
Not on my computer at the moment so I can't check the site's code, but it should be easily fixable by changing the cutoff point where the mobile menu is no longer used. iPad portrait tends to be 768px I think? If no one else gets to it first, I can look in to it over the weekend.

jospoortvliet:
@mikebabb if you have a device to test that probably helps, I don't :cry:

mikebabb:
@jospoortvliet @williambargent It even attempts a bit of touch emulation which can come in handy for testing touch device only enhancements. How effective it is I don't know, but for troubleshooting layout bugs like this one dev tools gets the job done

![screenshot from 2016-07-04 09 49 42](https://cloud.githubusercontent.com/assets/1648598/16555503/e36823f4-41cc-11e6-9ac3-cd153dd6fb98.png)

williambargent:
@mikebabb  I have been using it form a year or two and works perfectly. If I can test it on a physical device I usually do after.

PR #66, opened 2016-07-09
Fix 46, keep mobile menu until 1024px device width

mikebabb:
This is a fix for issue #46 that keeps the mobile menu button showing up until 1024px device width. Currently, it disappears and the site shows the "full" menu layout as of 768px, but as you can see from @mar1u5's screenshot the menu layout doesn't quite work at that size. 

I couldn't find the source file for the CSS initially and committed the fix in in the minified CSS file, which would have been overwritten as soon as somebody ran the grunt command to minify CSS again. Commit 	5a78685 corrects this by making the changes in navbar.less, where they should have been. Apologies for the mixup there.


jospoortvliet:
Pff, I only recently figured out grunt and nearly never run it. Best to make CSS changes in styles.css, that's where I do it ;-)

mikebabb:
In my day job I'm a pretty heavy grunt user, but I have a special hatred for bootstrap!

Does it particularly need updating? Should I apply this fix another way?

jospoortvliet:
I've only recently updated it so it is unlikely that it will happen any time soon - but, you know, some day... And then this fix would disappear ;-)

I think it's best to just put it in styles.css,  I know it is perhaps not super correct but it'll work and be retained in the future.

mikebabb:
Commit 4cddb73 adds menu CSS overrides to styles.css as requested. This way, updates to Bootstrap in future shouldn't cause the iPad portrait menu layout fix to be undone when navbar.less is updated.

jospoortvliet:
Awesome ;-) :+1:

jospoortvliet:
This change made the top bar move when items in the menu were hovered on my system, fixed in https://github.com/nextcloud/nextcloud.com/commit/49431b634931cd50468d5c5ad1bbf417b8634bff
(also re-ran grunt)

PR #67, opened 2016-07-11
Update about page with media stuff

jospoortvliet:
Add some links to interviews and articles about Nextcloud on our About page. What do you think and which important ones are missing? I don't want to create an exhaustive list but cover the important things...

about - nextcloud


jospoortvliet:
Whohoo, even managed to squash the shit out of this :dancer: 

@mar1u5 could you look at this, too? Sorry that it's a bit messy with whitespace... And I added a non-minimized version of the jquery youtubepopup as I had to make changes to that. And OMG, I am clueless... but I got it to work as I wanted and will just use it on the about page for now to avoid breaking other pages.

jospoortvliet:
It now looks better (and this PR will then make the videos on the features page also fit better):
about - nextcloud

The videos start at the time put in the id.

@jancborchardt like?

jancborchardt:
Awesome, I just wanted to ask if it’s possible to make the video part flush with the text. Looks great! :+1:

PR #88, opened 2016-07-25
add demo

jospoortvliet:
We now have a demo!

Also improved the calendar section a little.


jancborchardt:
Looks good. Fixed the alt text :+1:

PR #89, opened 2016-07-26
move demo next to Download action

jancborchardt:
Makes a bit more sense to have it next to »Download« instead of all the way on the left. @jospoortvliet @nextcloud/designers


mar1u5:
Makes sense :+1:

Bugsbane:
As Ma1u5 said, it makes sense to have it just before Download, so if people are hesitating to get started, then clicking on and seeing the Demo can make them feel more comfortable first. :+1:

PR #102, opened 2016-08-09
Improve "Get Involved" page

mar1u5:
Some small fixes to our "Get Involved" page. We are not longer bootstrapping our project, we are calling it server instead of core, and so on 😜 

cc @jospoortvliet


jospoortvliet:
awesome stuff, thanks!

jospoortvliet:
going live :D

PR #104, opened 2016-08-16
Changed faulty code

ezraholm50:
Said docker pull docker pull, should be docker pull collabora/code IMO. Tested the faulty one and it doesn't work.


jospoortvliet:
sigh, my bad... thanks!

PR #109, opened 2016-08-19
AutoLogin via Script

Sen5ation:
Hi,

i'm wondering if there is or should be a way to login via script? Or like an api-call?
Is there something like that? Did i miss it? 
I want to do a SingleSignOn-Script for some apps including nexcloud.

Whould be great to hear some advices.

Grz.


MariusBluem:
Can anybody explain me what happens in this PR or what is it for 😱😜 @jospoortvliet

PR #121, opened 2016-08-30
add 3rd party section on install

jospoortvliet:
So this adds three other compatible mobile clients to our install page. Are there more?


jospoortvliet:
tested, works...

spectacle z13750

PR #126, opened 2016-09-07
Remove AppStore Buttons / Fix Styling of the Buttons

MariusBluem:
I think we should not show the AppStore Buttons directly on the frontpage, because we have also some other clients (desktop) ... Before the fork, we have linked with the buttons to links (Desktop/Mobile) ... This is also fixing the font of the buttons, which is currently bigger in "2. Sync your data" then "1. Get Nextcloud Server" 😉

"More Clients" should be "Mobile Clients" ;)

We have now one appliance, so we should mention that in the texts 😁 

cc @jospoortvliet

### DIFF
--- a/page-install.php
+++ b/page-install.php
@@ -9,7 +9,7 @@
 			<i class="icon-download"></i>
 		</div>
 		<h2>1. Get Nextcloud Server</h2>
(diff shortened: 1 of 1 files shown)


jospoortvliet:
true. Let's go ;-)

jospoortvliet:
thanks!!!

PR #130, opened 2016-09-08
Installation Page text a bit longer so that it will better fit with the 1. step (2 lines)

MariusBluem:
Currently, the text under the 2. step is a bit to short so the buttons are on different positions 😁 
... this will fix that 😉 

cc @jospoortvliet


jospoortvliet:
:+1: thanks for your attention to details ;-)

PR #134, opened 2016-09-14
fix typos and social event date

jancborchardt:
Please check @Jospoortvliet – I assume the social event is the Schleusenkrug-Dinner?


jospoortvliet:
Thanks Jan, you're right ;-)

jancborchardt:
> Thanks Jan, you're right ;-)

Going to get that printed and framed

PR #138, opened 2016-09-17
add note about hiring

jospoortvliet:
@jancborchardt @juliushaertl @karlitschek what do you think about adding a link to our jobs page on our contribute page, with this:

> Note that we're hiring. We look especially for people who want to turn their passion into their job!
> When possible, we hire people from our active community base or from the wider open source community.

Help me find better words ;-)


karlitschek:
i like the idea

PR #139, opened 2016-09-17
List Devices (Nextcloud Box)

MariusBluem:
- Added Nextcloud Box as a "Get Nextcloud Server" method 
- Fixed confusion about direct download and detailed download-button by simply merging them into one
- Better spelling for the text

cc @jospoortvliet


MariusBluem:
**WIP**

MariusBluem:
Now the text should be short enough to not break the style 😅 Ready for a review @jospoortvliet

jospoortvliet:
@MariusBluem so yes on 90%, but I think we should not link "devices" to "Box" but have a page like the providers page so we can link to the Spreedbox, the Nextcloud Box and other devices & how to create your own. Want to give creation of such a page a shot? ;-)

MariusBluem:
@jospoortvliet agreed and this would be also the plan, but i think it is currently out of scope ;) ...spreedbox is not orderable (out of Stock) and there is no other Hardware i know about we could list on such a page 😁

jospoortvliet:
@MariusBluem actually... there's a company in New York which makes something... @blizzz said he knows who they are :dancer: 
And the Spreedbox is not out of stock for enterprise users so we should at least link to it.

But more importantly, if this link links to this one device, people might get the impression that this is what we do. So even if we did only this one box, we would still have to have a page which explains we want others to do things like this - we are no hardware company and don't want to be making consumer devices...

blizzz:
@MariusBluem @jospoortvliet http://syncloud.org/ are working on a release with Nextcloud

jospoortvliet:
@MariusBluem so I created a 'devices' page with the Spreedbox and the Box. There'll be other hardware for sure, when we advertise it ppl will probably tell us ;-)

You OK with this?

MariusBluem:
Great :+1: Can we merge this?

jospoortvliet:
will merge.

jospoortvliet:
done and live. Thanks!!! See https://nextcloud.com/devices ;-)

MariusBluem:
Uhm ... the Nextcloud Box is looking strange there (too high) ...

jospoortvliet:
should be fixed.

PR #166, opened 2016-10-07
Remove duplicate entry

MorrisJobke:
compare with line 64 ;)

cc @Espina2 @LukasReschke


MorrisJobke:
And this fixes the text on the buttons of the enterprise page. It's now better centered.

LukasReschke:
THX! Deployed.

Espina2:
Don't remove the padding value in all the buttons the visual height of the arrow have to be compensated with less margin in the left. The label should not be centered so the button doesnt look centered.

Espina2:
Ah forget its something that repeated. :) Sorry

PR #168, opened 2016-10-08
remove blur effect from icon, compress icon files

jancborchardt:
Please review @nextcloud/designers @LukasReschke


skjnldsv:
```
19:47 [xxxxxx@xxxxxxxx] ~% svgo logo_nextcloud_blue.svg 
Done in 128 ms!
5.179 KiB - 20% = 4.146 KiB

20:18 [xxxxxx@xxxxxxxx] ~% svgo logo_nextcloud_white.svg 
Done in 115 ms!
4.985 KiB - 19.9% = 3.992 KiB
```
We can go deeper! 😜

jancborchardt:
@skjnldsv the problem with svgo is as said before that it’s heavily lossy and can lead to issues with the images. :\ I only always use the scour script in server/core/img/image-optimization.sh

skjnldsv:
@jancborchardt do you have an example? I never encountered an issue like that. :smile:
The only loss you can get is an information loss (only the useless ones). Svgo is designed to prevent any render loss on display :confused:

jancborchardt:
@skjnldsv yeah, we had a discussion on it, maybe regarding some app icons with @Bugsbane or @juliushaertl? Not sure but I think we had some lossy results with svgo, and scour was the better choice.

juliushaertl:
Not sure about that, when I was committing svg stuff, svgo has been recommended to me.

We should add some information on how to handle SVG optimization to the docs.

jancborchardt:
@juliushaertl agree we should have it in the docs. So far the core/img/image-optimization.sh file is the one I used. Or at least the scour line for SVG:
```
scour -i $svg.opttmp -o $svg --create-groups --enable-id-stripping --enable-comment-stripping --shorten-ids --remove-metadata;
```
And then Trimage for PNG & JPG: https://trimage.org/

jancborchardt:
@juliushaertl agree we should have it in the docs (cc @eppfel ;). So far the core/img/image-optimization.sh file is the one I used. Or at least the scour line for SVG:
```
scour -i $svg.opttmp -o $svg --create-groups --enable-id-stripping --enable-comment-stripping --shorten-ids --remove-metadata;
```
And then Trimage for PNG & JPG: https://trimage.org/

PR #169, opened 2016-10-08
use dash instead of pipe in title, as we do

jancborchardt:
@LukasReschke right? ;D


LukasReschke:
Also head-old.php otherwise that will only affect new files.

(basically redesigned pages use head.php, old files the others)

On my mobile atm otherwise I'd have changed this myself :)

jancborchardt:
@LukasReschke added :)

PR #171, opened 2016-10-10
Added Animations on HomePage and Enterprise

Espina2:


Espina2:
@nextcloud/designers @LukasReschke can review and merge?

jancborchardt:
@LukasReschke can you deploy it to staging?

assets/css/custom.css
```diff
@@ -407,8 +407,20 @@ section#why-nextcloud {
       padding-right: 30px; }
     .Header-Navigation .navbar li a {
       color: white; }
+      .Header-Navigation .navbar li a:after {
```
skjnldsv: This is breaking the download button! :/
screen

Espina2: This is added in the last commit ( I shouldn't commit that yet). I trying to figure out why a not :last-child doesn't work for an after pseudo-element.

Espina2: But are editing directly the .css file, I'm talking about the .scss file. But if made it working just made a commit with the last-child not having an after element. This should work

skjnldsv: Okay! Good luck. The rest looks good :)

LukasReschke:
Resolved the conflict and deployed at https://nextcloud.com/enterprise/?staging=true and https://nextcloud.com/?staging=true

Let me see how to fix the download button…

assets/css/enterprise.scss
```diff
@@ -40,20 +40,23 @@
...
-        height: 60px;
-        padding-top: 16px
+    
+	  .topheader {
+      	padding: 0;
+      
+		a{
+			width: 290px;
+			height: 60px;
+			padding-top: 16px
```
LukasReschke: This looks wrongly intended.

assets/css/enterprise.scss
```diff
@@ -40,20 +40,23 @@
...
       }
     }
   }
 
   .container-fluid{
-    img {
-      height: 100%;
-      width: 100%;
+    
+	  img {
```
LukasReschke: As well. Might want to look into caps vs. spaces in your text editing tool :)

Espina2:
I think you not include the main.js function in the deploy. Can you check @LukasReschke

Espina2:
@LukasReschke @jospoortvliet We need to deploy the main.js for testing.

LukasReschke:
> @LukasReschke @jospoortvliet We need to deploy the main.js for testing.

It is. But it doesn't seem to be included anywhere or am I missing something? :)

Espina2:
You have misses something. Maybe some script code in the homepage html. @LukasReschke can you take a look?

LukasReschke:
> You have misses something. Maybe some script code in the homepage html. @LukasReschke can you take a look?

Sure. So "main.js" should be included on every page or just on a specific one?

Espina2:
Yup, for now the main.js should be in everypage. More later we need to think about require.js.

jospoortvliet:
2 questions/comments.
1. can we do this in css, not javascript and
2. I note that the texts under the three items just under the header on the front page has disappeared.

PR #177, opened 2016-10-11
Added another possible solution for collabora problems

jakob42:
Only added simple HTML including a solution to problems with the collabora app. And added a missing `<code>`-tag.


jospoortvliet:
Thanks, @jakob42 for your contribution helping ppl troubleshoot problems! This makes a big difference and I appreciate it, as I'm sure will many users.

jakob42:
Thanks for all the work on the amazing software. Just doing a small part when I'm able to. :)

PR #181, opened 2016-10-13
Changes on the docker command documentation

jakob42:
Info on how to use one docker container with multiple nextcloud installations

Also changed the syntax to create the docker container. The original syntax did work, but the backslash to escape the dot gets eaten by bash and the result in /etc/loolwsd/loolwsd.xml
in the container was potentially insecure, because the (unescaped) dots would have allowed access to the domain cloudxnextcloud.com as well.

On a side note: Is a pull request the preferred way to change documentation or should I open issues?


jospoortvliet:
Awesome improvements.

yeah, this is the right way. At some point we should look into putting much of this in the actual documentation, perhaps, but it is quite a 'special' thing so it makes sense to keep it here, too.

Thanks for these changes, really appreciated!

PR #187, opened 2016-10-16
Update page-webrtc.php

devz3r0:
the old version does not work, because:
1. docker creates a own network interface, so it listens only in the container itself.
2. no mapped directories to the conf file and the nextcloud app.
3. limit access via docker run command.
4. workaround for virtual machines.


jospoortvliet:
Wow, you're on a roll. Great work. I haven't tested it but I have no problem believing you - I'll merge it!

Thanks again!

PR #205, opened 2016-10-18
Optimg

jospoortvliet:
both the arrows became black and down-pointing after optimization. Un-optimized them...


jancborchardt:
And this, dear @jospoortvliet @nextcloud/designers @skjnldsv, is why we do **not** use optipng. :)
Instead, please use Scour as it is less lossy and does not fuck up the SVGs completely, which are our source files too after all: https://github.com/nextcloud/server/blob/master/core/img/image-optimization.sh#L10

skjnldsv:
You mean optipng or svgo?
Could you five me an example of a failed svg? I want to investigate why svgo is a problem! 😝

jancborchardt:
Sorry – svgo I guess. @jospoortvliet that’s what you used, right?

LukasReschke:
> Could you five me an example of a failed svg? I want to investigate why svgo is a problem! 😝

I guess just looking at the tab "Files changed" helps here. 😉 

https://github.com/nextcloud/nextcloud.com/blob/f396210e3cd968a5ac7afaf45865771de2374c9f/assets/img/arrow_right.svg
https://github.com/nextcloud/nextcloud.com/blob/32806526260eaed08f77fe6674114dee151e6c23/assets/img/arrow_right.svg

Completely broken…

skjnldsv:
I can't load the files tab, chromium can't keep up :p
Thanks Lukas!

skjnldsv:
This isn't due to svgo. The svg file is incorrect. Even inkscape can't display it correctly! :/ 
Like I though, svgo only mess up corrupted svg. If you correctly make this svg and save it with inkscape, svgo works perfectly fine :)

Espina2:
Inkscape its not the only svg software, It should work with illustrator, sketch, affinity, corel etc, etc.

jancborchardt:
@Espina2 that’s not what @skjnldsv said. ;) But the SVG file which was exported by Illustrator (I guess) seems broken. Stuff like that didn’t happen so far with the files exported with Inkscape. How can we solve that, maybe there’s a setting when saving it in Illustrator?

Espina2:
This is weird, maybe they are exported from zeplin are o sketch. I never had problems with illustrator. But lets test what is the problem. I will post the original file and you guys make the test.

skjnldsv:
But @jancborchardt is right though just use scour, I just don't understand why svgo is a problem since it's beeing used by so many people out there! 😕 

Anyhow, don't worry @Espina2, I just meant that opening your svg into inkscape and degrouping it solves the problem. I don't know why though. Let's not take too much of our precious time just for some svg!: ;)

Thanks for your work!!

jospoortvliet:
We should be careful in the future not to run this over all files. For now, we've saved a lot of space with this so let's just be happy :D

PR #206, opened 2016-10-18
enterprise page changes

jospoortvliet:
Some changes to how we display pricing & other changes requested by sales.

Some open questions/issues:
* [ ] the new mouse-over effect (replacing tooltips) has z-index issues with the lines on the bottom (just above support)
* [ ] how should the price section look? Also with th checkmarks, or nothing, or something else?
* [x] I'd like to make some of the items links. How should that look, dotted line under the text I suppose?

Note that I'll bring back the arrows in the FAQ section in another PR.

This is what it looks like now. @jancborchardt @Espina2 @MariusBluem thoughts?

enterprise nextcloud


jospoortvliet:
deployed to staging:
https://nextcloud.com/enterprise/?staging=true

Ignore the FAQ section, it has some errors fixed in master. Look at the table and its mouse-over effects!

Espina2:
@jospoortvliet 
screen shot 2016-10-18 at 20 41 57
Why I have a different huge arrow in all buttons now? 

The tables are not designed to handle so many information. But I think you can remove all the icon and only use the grey color for what the different versions dont have. This should give more space to the text. 

If you have links you should use the default style, blue and underline. 

> the new mouse-over effect (replacing tooltips) has z-index issues with the lines on the bottom (just above support) This part I don't get it

jospoortvliet:
Let me see if I can shorten the tables a little, remove some thing...

Will use the default style for the links then. But most items in the table will be links... hope it won't look bad.

The arrow is probably due to some css that applies to all pages. You don't want it?

> This part I don't get it
screenshot_20161019_101958

Espina2:
I like the default arrow I dont know who changed it but its not me. 

What I think is if transform everything in links you will loose the attention of the users. Because you are opening new pages for them look, to go back to open another one. 

You can have maybe a button to features, where you show all of them. 

I don't believe that a "wikipedia article page" is good to this. This should be fast and very easy to understand... The tooltips I think its a good ideia but if they existe we have to have the little _i_ icon.

jancborchardt:
Yeah – the arrow should be the same as on the frontpage for sure.

I also agree that the table should be concise. But depends on what sales says converts …

Also capitalization: »Plans and pricing«, »Enterprise support subscriptions« – only capitalize the first word.

jospoortvliet:
Fixed capitalization.

I have only three links now: migration and WebRTC/Collabora Online.

I shortened the list by removing:
* Mobile and desktop clients
* Enterprise features
* SIP gateway

Tooltips are still broken, any idea how I can fix them?

jancborchardt:
@jospoortvliet the tooltip thing is a z-index problem. You need to add `z-index: 10;` to the `.overinfo .infotext` rule.

jospoortvliet:
@jancborchardt can you make it work in the browser? I can't. Besides, z-index: 10; is already there in that class ;-)

jancborchardt:
Hm, it worked in the browser. I assume you tried adding `!important` already too, and increasing the number? Not sure what the issue is then.

jospoortvliet:
@jancborchardt show me  a screenshot of where you added it. I haven't been able to add it anywhere in the browser where it actually worked. Still overlap over the buttons on the bottom etc etc.

Espina2:
I can't make it work to. But the line have some weird negative z-index. Maybe its related

jospoortvliet:
Not sure, the buttons on the bottom also overlap.

jancborchardt:
Hm, I can’t make it work anymore either … maybe it has to do with the animations of the elements coming in?

Espina2:
Maybe but this are easy to test. Just remove the class revealOnScroll and test if everything is okay.

jospoortvliet:
done in other PR's.

PR #211, opened 2016-10-19
[RFC] Switched Home and Bussiness Image

Espina2:
https://github.com/nextcloud/nextcloud.com/issues/197

Ignore this two commits in the merge  [WIP] Initial Structure [WIP] Finished the structure

PR #213, opened 2016-10-20
Homepage img

jospoortvliet:
replacement for #211 with only the changes needed ;-)


Espina2:
Awesome. @oparoz  @jancborchardt  can you check if the image are okay?

oparoz:
Better than before :) :+1:

jancborchardt:
Good stuff, text is now more readable too :+1:

PR #217, opened 2016-10-22
linkify some items in the list

jospoortvliet:
Make it easier for potential customers to figure out what Collabora is or what migration we offer.
screenshot_20161022_102058


oparoz:
:+1:

Espina2:
I dont really like to have this repeated links. We should in another solution for display this content.

jospoortvliet:
@Espina2 well the original design had all the items on the left and just check boxes in the columns, but you are the one who changed that into three columns which have each a copy of each others' info ;-)

Espina2:
@jospoortvliet yup, its right but the other solution doesn't work in mobile phones at all. And this is the problem to not have all the content before design. You design with one think in consideration and when you add more things it breaks. Its still "works" but looks ugly and have very repeated content. 

If you see here the same solution works fine for the same thing. https://evernote.com/intl/pt/pricing/

jospoortvliet:
yeah, it works, and honestly I think we're good. Nothing is ever perfect but it is quite ok.

oparoz:
It sorts of works for Evernote because they have exactly the same content on the same line, but we don't, so our presentation is confusing.

Ideally we would have columns for mobile and the table for desktop...

Espina2:
@jospoortvliet for sure, and this make the work, my eyes just dont like to see all this repeated content and "links" everywhere. For now its perfect don't get me wrong, just pointing that it can be improved. But I will think about it, and I have sure that I can point a better solution. :)

@oparoz maybe you are right. 

Overall this page is improved a lot from the previous page. 👍

karlitschek:
👍

jospoortvliet:
I'll get this in then.

PR #221, opened 2016-10-22
It's nextcloud.com

MariusBluem:
cc @jospoortvliet


jospoortvliet:
so it might be github but if I look at the diff I see NO difference at all! What did you change?

MariusBluem:
You are right ... GitHub is not showing. But if you compare the `master`-page with the page from this branch, you will see that i have changed **nextcloud.org/news** into **nextcloud.com/news** 😉

jospoortvliet:
... and that's a good move :D

PR #223, opened 2016-10-22
shorten list by removing some items

jospoortvliet:
I shortened the list by removing:
*    Mobile and desktop clients
*    Enterprise features
*    SIP gateway


jospoortvliet:
We'll find better ways to do this...

PR #233, opened 2016-10-23
[RFC] Added javascript vendor files

Espina2:


jospoortvliet:
Ok, works here. But it is bad we add even more javascript to the head, it has to be put in the footer and the website should not be broken without javascript as it is now. I'll create a issue for that.

Espina2:
Just pay attention to the depencies. If something doesn't load when its needed it can break the page.

PR #235, opened 2016-10-24
Improve Client Downloads

MariusBluem:
* Recommending CryptoCloud does not makes any sense (since we have the Nextcloud iOS by CryptoCloud)
* DavDroid does also sync Tasks
* Improve WebDAV text, and also implement it for desktop clients - section
* Move CrossCloud to desktop-section, because its a desktop client 
* Move sources and APK download directly under the android stores, because otherwise, it would look like we have sources for iOS :)

cc @jospoortvliet

### DIFF
--- a/install-instructions.php
+++ b/install-instructions.php
@@ -103,7 +103,7 @@
 				</div>
 	</div>
 <!-- </div> -->
-<!--  Mask instructions server -->	
+<!--  Mask instructions server -->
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Sorry it took me so long to review but - awesome work. Tested, works, merging!

jospoortvliet:
and live...

PR #242, opened 2016-10-27
Jobs

jospoortvliet:
Adding 2 jobs: sales engineer and account manager.

jobs nextcloud


Espina2:
Can I suggest to collapse all the information of the "job" and only show them when toggled?

jospoortvliet:
Yeah... yeah, I can do that.

jospoortvliet:
spectacle j32444

So now it is collapsible and I 'upgraded' to the new style because I had issues with the collapse (somehow bootstrap is loaded twice which breaks collapsible links)

Sadly, the reveal-on-scroll doesn't work when a page is so big you can't scroll :disappointed: 

what do I do about that? just disable it in the footer?

Espina2:
Okey, tomorrow I will take a look with you in this page. Maybe add a nice image on top like the new pages, and some tweak here and here.

PR #252, opened 2016-11-04
page-security.php spelling fix

reillychase:


jospoortvliet:
Sjees, diff totally fails to show what changed... Probably due to the stupid newline in the php code for the translations, sorry about that.

But - good catch ;-)

PR #257, opened 2016-11-08
Fix HTML layout

nickvergessen:
Also there was a comment on IRC:

```irc
<strk> this page reports the android client as being "beta", but on f-droind I find both a "beta" and a non-"beta" version: https://nextcloud.com/install/#install-clients
<strk> the non-beta being 1.3.1, added on late September 2016
<strk> was it a mistake by f-droid maintainers to publish it as non-beta or does the webpage need be updated ?
```

So the page seems to need some rewording to make clear that the big buttons are **not** beta.

@jospoortvliet @AndyScherzinger

### DIFF
--- a/install-instructions.php
+++ b/install-instructions.php
@@ -13,7 +13,7 @@
 							<p><?php echo $l->t('Follow the');?> <a href="<?php echo $DOCUMENTATION_ADMIN; ?>installation"><?php echo $l->t('Nextcloud Admin Manuals</a> installation chapter.</br>
 							If you already run Nextcloud, refer to the');?> <a href="<?php echo $DOCUMENTATION_ADMIN; ?>maintenance/upgrade.html"><?php echo $l->t('upgrade manual.');?></a><br/>
 							<?php echo $l->t('<a href="/enterprise">Need enterprise support?</a>');?></p>
-							
+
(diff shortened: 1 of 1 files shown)


MariusBluem:
Beta links should be moved to **Get Involved** section of Download-page because this is the place where you can find the server's daily builds too 😉

nickvergessen:
Issue has now been reported in https://github.com/nextcloud/nextcloud.com/issues/258

AndyScherzinger:
We actually have 3 different client releases:

* Stable (published on Google Play Store and f-droid)
* Release candidates (published via Google Play Store Beta Programm, same app as stable just a newer version)
* Beta (only published on f-droid), actually a different app, which can be installed next to stable/RC and is not a pre-release of the next version but a container for _all_ upcoming and implemented features and for testing purposes.

So like @MariusBluem said, the "real" beta should probably be in *Get Involved* since it is for testing, really early adopters, basically people who like to help us out with testing new stuff and testing on a high number of environment setups, devices. :)

MariusBluem:
:+1:

jospoortvliet:
Thanks for the fixes here. What do you think about moving them to the get involved section: https://github.com/nextcloud/nextcloud.com/pull/265

AndyScherzinger:
Sounds good to me and also to be the more adequate place :+1:

PR #266, opened 2016-11-11
Slideshow

Espina2:
So I fix the "Slideshow" on the homepage. Now it changes the image correctly without completely hijack the scroll. 

Thing not fixed yet. 

- [x] Create function to animate automatically all the images without creating a new scene. 

- [ ] Add active state and anchor point to indicators

- [ ] Mobile

I'm not quite sure if you can preview this here because this wordpress themes have a lot of hacks
cloud.paulomoura.com.pt/wordpress

@nextcloud/designers @LukasReschke


Espina2:
@ChristophWurst 

Jos said that you want to help me with javascript and front-end stuff. Do you think you see if you can finish this?

ChristophWurst:
I can take a look at this next week

Espina2:
Nice, let me know if need help. I will not have many time this week to.

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/.idea/next.iml" filepath="$PROJECT_DIR$/.idea/next.iml" />
```
ChristophWurst: do we really need/want all those intellij files in here? cc @jospoortvliet

Espina2: This is my fault. They are files from visual studio code. I should out them in the gitignore

jospoortvliet:
@Espina2 can you be specific about what you need from @ChristophWurst here?

Espina2:
Yup of course. So I need help in add active state of the indicators. (image above) 
screen shot 2016-11-28 at 14 54 58

And in mobile we need to make a swipe "carousel" like I designed in the link above.
https://scene.zeplin.io/project/57bf54c24ff046f70c9b9694/screen/57c9a9f78188415217a50e93

ChristophWurst:
tlocal

> below indicators.

How do we generate those indicators? I see that you've used some plugins like 'ScrollMagic'. Is there any  other magic involved in the slideshow/scroll magic? This is all very new to me, so I have to dive into some documentation before I can understand the code.

ChristophWurst:
> Conflicting files
> .idea/workspace.xml

This shouldn't be in the repo. Apparently it was commited in another branch/PR. @Espina2 could you please remove the `.idea/*` files? Thanks :-)

Espina2:
Yes of course I can remove that. I will take a look at that hidden files and remove them. You should see how you can catch/trigger a position of the element and addClass .active to the indicator. So the first text should addClass .active to the first indicator, the second should add the class to the second and remove from the previous one. You can do that with jquery I guess, you just need to know what text are in the window and addClass to the indicator. 

If you dont understand, fell free to add me on skype if you want.

ChristophWurst:
> Yes of course I can remove that. I will take a look at that hidden files and remove them. You should see how you can catch/trigger a position of the element and addClass .active to the indicator. So the first text should addClass .active to the first indicator, the second should add the class to the second and remove from the previous one. You can do that with jquery I guess, you just need to know what text are in the window and addClass to the indicator.

Done. Apparently that has no visual effect. The class is now added/removed automatically. Do you know how I have to adapt that indicator to change its color or whatever?

Espina2:
You are targeting the <a> instead of the li with the class "btn_carousel". Just need to target the correct thing. :)

Espina2:
I see that you are using php to get the job done, so I can't understand very good what you are doing. ;)

ChristophWurst:
> I see that you are using php to get the job done, so I can't understand very good what you are doing. ;)

I'm not 😉

Espina2:
Yup you are right. You are using const instead of var and this messed up my brain. :)

ChristophWurst:
> You are targeting the <a instead of the li with the class "btn_carousel". Just need to target the correct thing. :)

Whoops. Fixed. Now it's somewhat working, but the "enter" and "leave" events are not reliable, e.g. when I scroll down to the bottom of the page the "leave" event is not triggered, when I scroll up it is. 😕 Do you have experience with that library? Are we using it wrongly or is that the expected functionality?

Espina2:
I have some experience, but I am not an expert. The behaviour expected is like this example https://www.uber.com/ .

Now that you said that, I think you only have the solution to create a scene for all the sections and give to the image a specific position instead of adding more x pixels. 

I tried to not do that way because this will add more code and I think that should be another way to do it.

Espina2:
And for the problem that you have with all the active states, you just need to remove all of them before you add the active. Instead of having this in the leave "$indicator.removeClass('active');" you have this in         $(".btn_carousel").removeClass("active"); inside of the each.

ChristophWurst:
> And for the problem that you have with all the active states, you just need to remove all of them before you add the active. Instead of having this in the leave "$indicator.removeClass('active');" you have this in $(".btn_carousel").removeClass("active"); inside of the each.

Makes sens. I've pushed another commit. However, this does only work when scrolling down. No idea why, but if you scroll up it does not fire that "enter" callback.

Espina2:
I think something like this can resolve all our problems and make this a reusable component. 
https://github.com/protonet/jquery.inview

Basically what Im thinking about is adding unique id to each "section" and if that section are inview they trigger the active class and remove all the others, and instead of adding/ removing more pixels to the image we can say something like this " if section 2 is inView toggleClass"image-position-2".

ChristophWurst:
> I think something like this can resolve all our problems and make this a reusable component. ( I never used)
> https://github.com/protonet/jquery.inview

Do we really want to include yet another library? What are the benefits?

Please elaborate what you did in https://github.com/nextcloud/nextcloud.com/pull/266/files#diff-f8d30c93d174d13f3976c4c97a8e0ad2R40. I don't understand most the code. Could you please add some meaningful comments? :)

Espina2:
I want to make a component that we can reuse in the website. Like I write we can make this  single slide work pretty easy you create a different scene for every textTrigger elements and give different instructions for everyone. The challenge here is, that component detects how many elements that we have and just to the magic, instead of declaring every single time that we use them.  

This plugin that I said have only 150 lines of code, and I think that we can help us set different actions by looking all the textTrigger are inView and define what to do in that specific case. 

But feel free to that how you want, like I said im not a developer, just a curious designer.

Espina2:
And please don't use constants in javascript they are more slow then var. You can read why here http://stackoverflow.com/questions/33295772/does-it-make-sense-to-use-const-for-all-variables-that-will-never-be-changed

ChristophWurst:
> And please don't use constants in javascript they are more slow then var. You can read why here http://stackoverflow.com/questions/33295772/does-it-make-sense-to-use-const-for-all-variables-that-will-never-be-changed

I have to disagree here. Read the SO answer again. It was slower years ago but now it's at least as fast, on optimized js engines it's even faster 😉 
In any case, I'm also fine with reverting to vars. There's much other code that is slow on our page. It won't make a difference.

ChristophWurst:
Could you please clarify what a "reusable component" should look like? I have no experience with this kind of js hackery and all the libs you're using here. It makes also no sense to replace those with others because I'd have to learn others too. This is all very new and confusing to me.
To be able to help you here I need some clear problems to solve. From reading the code I have the feeling that different, unrelated parts of the website have been touched. I can not reconstruct what you've done so far in this PR unfortunately.
Would it make sense to start from scratch? E.g. start a new branch based on the current master?

Espina2:
So the problem is that we want to use a slideshow in other pages without write more javascript code to each different page that we exist. This is what we want to do. This are the problem here. 

I don't think you need to create another PR but if you think its okay to me. 

On the code we are only targeting the homepage and beside of the slideshow you are calling some SVG animations. Sorry but Im not getting what you dont understand. 

The conflicts are just maps from the scss files.

Please add me on Skype (expina21) its more easy to talk and see what you don't understand since you can show printscreens, and how I can clarify better how I pretend.

ChristophWurst:
Hm, seems like your last commit made the diff even bigger and removed the stuff that I've worked on so far. May I remove that commit again?
I don't use Skype, but I'd be happy if we can discuss this on IRC :-)

ChristophWurst:
Do we have a state/commit where the slideshow actually worked?

jospoortvliet:
so all in all, this is done, correct, and can be closed? @Espina2

Espina2:
This still have the HTML needed to the slideshow on homepage works as in the enterprise. 

So maybe you can copy the HTML to a new branch, test and if it works as expected merge with master and deploy.

@jospoortvliet

PR #268, opened 2016-11-11
Remove Windows Mobile from Enterprise

MariusBluem:
Nextcloud for Windows Mobile is very cool and I love the development I can see going on that ... But nothing Enterprise seems to support currently. We can consider about that later, if the app is more tested and ready to be supported by Nextcloud GmbH, which has to take responsibility to (security) issues etc. - which is out of scope for now 😁 

cc @SunboX @jospoortvliet @LukasReschke @rullzer


jospoortvliet:
thanks dude, I should not have added this, it's not even announced..

PR #274, opened 2016-11-25
[WIP/NI] Announcements highlight in the Homepage

Espina2:
screenshot_20161125_194005

So after I talk with @jospoortvliet  about how we can highlight some things that are new I came out with this solution. 

Since @jancborchardt don't like carousels (and we have some valid points) I think its the best way to do it. 

@nextcloud/designers


page-homepage.php
```diff
@@ -6,6 +6,10 @@
 <!-- 	<div class="container-fluid"> -->
 		<div class="row">
 			<div class="col-md-6 toptext ">
+				<a class="annoucement" href="#">
+				<span class="type">Fresh</span>
+				<span class="message"><strong>Nextcloud 11 has just arrive.</strong> Take a look at the new features</span>
```
tcitworld: arrived

Espina2: The text is dummy, so it doesn't matter at this point. :)

jospoortvliet:
I like it! Absolutely works for me. Really cool, actually. If Jan agrees, we should merge it then just hide it until we need it.

Espina2:
Yes you can comment the code and use it only when it needed. But for example you can already highlight the beta version. :)

jancborchardt:
@Espina2 just one thing: Can you put it _below_ the »Get Nextcloud« button? Nothing should ever be above the slogan and call to action.

jancborchardt:
(Btw the animated logo looks cool as well! :) cc @jospoortvliet all my mail apps have issues these days for some reason, so here’s the reply ;)

Espina2:
I attach the image like you want. I didn't like it, it makes this not work good in mobile devices, and dont look anymore like an highlight. 

screenshot_20161130_115900
screenshot_20161130_115934

jancborchardt:
@Espina2 to fix the layout issue on mobile, put the »Fresh« tag in a separate line above maybe :)

> I didn't get it why it should be not above the slogan

Because it’s information overload. It almost looks like a banner ad, which people are prone to ignore. Nothing should take attention away from the slogan, the short description and the call to action. That’s the most important thing on the page.

Espina2:
> Because it’s information overload. It almost looks like a banner ad, which people are prone to ignore. 

Pushing the thing above doesn't change this. 

I don't agree with you, having this on top just look something new are come, this is the purpose of having this, doesn't make any sense if this are the last thing you are reading. I prefer to have on top or not have it all. 

Since moving this above is not my vision and doesn't represent what I think is best, feel free to change yourself or erase this PR.

jospoortvliet:
Honestly I agree with @Espina2 here, if only because it simply looks better above than below... And while I get that it might distract from the main message, well, it is news, fresher... I think it is fine this way and could be merged. Jan, can you agree to disagree?

jancborchardt:
Ok whatever, go ahead.

cc @nextcloud/designers what do you think?

Espina2:
Let me just give a quick adjustment in mobile.

jospoortvliet:
@Espina2 you can merge once OK.

Espina2:
Merged

Espina2:
@jospoortvliet don't forget to change the link and the text. :)

PR #280, opened 2016-11-30
add menu items

jospoortvliet:
@Espina2 here's some more.


Espina2:
I discover another bug in Header. At 1200px the container breaks. I will fix that in the cross-browser PR.

Espina2:
I think I now fix all the bugs. You can merge this when is ready. :)

PR #283, opened 2016-12-01
Refactor css

jospoortvliet:
Clean up the CSS (get the 'background' stuff separated) and update the migration and contribute pages to the new style. Not individually designed and optimized, but good enough for now I think.

@Espina2 what do you think? Check it out, if it is OK, just merge pls...


jospoortvliet:
Ok, it works, checked most pages on our site - let's go...

PR #301, opened 2016-12-09
fix ghost button with arrow

Espina2:
@jospoortvliet 

Fix the arrow in buttons with transparent background.


jospoortvliet:
:+1:

PR #308, opened 2016-12-13
Make .nav bg-animation working in both directions

juliusberger:
The transition attribute of the nav is just applied, if .scrolledis added while scrolling down from the top of the page on desktop devices. 
A more consistent look would be achieved if the transition attribute belongs to .nav and not only to .scrolled

@jospoortvliet

### DIFF
--- a/assets/css/header.scss
+++ b/assets/css/header.scss
@@ -526,6 +526,8 @@
 
 .nav {
     padding-bottom: 70px;
+	
+    transition: all .2s ease-in;
 
     @media (max-width: 992px) {
         padding-bottom: 20px; 
@@ -533,7 +535,7 @@
 
     &.scrolled {
         position: fixed;
-        transition: all .2s ease-in;
+        
         background: white;
         box-shadow: 0px 1px 2px 0px rgba(0, 0, 0, 0.09);
 


Espina2:
@juliusberger what you made doesn't change nothing visual, its the same thing as it is now.

juliusberger:
@Espina2 I think it does make a change. Because like that, the transition attribute is applied in general to the .nav, and not only to .scrolled. For .nav__label it is already properly. It's a very minor visual difference, but it is one...

Espina2:
No, it doesn't. You can toggle the class in the inspector and you can see that this doesn't change anything. And that transition is only needed when you have the scrolled class. :)

Espina2:
@jospoortvliet you can merge this if you want, it doesn't change anything. :)

juliusberger:
I created a [JSFiddle](https://jsfiddle.net/1z8nwc1t/) to demonstrate what I mean... It's the difference while going back to white. The first div is like it is now, and the second one how I thought it should be...

jospoortvliet:
to me the jsfiddle is convincing... ;-)

Espina2:
So feel free to merge. :)

Issue #271, opened 2016-11-21
Remove support@nextcloud.com from legal notice


LukasReschke:
@jospoortvliet Please remove the support@nextcloud.com address for https://nextcloud.com/impressum/. I don't care what address we use there but not that one ;-)

Also that gets actually customer inquiries. So that should go to something that is read. But ending it up in our support tracker is very suboptimal. This will file tickets that all subscribed engineers will receive 🙈 

So we could setup a contact@ or whatever we like.

LukasReschke:
Also other occurrences may need to be adjusted: https://github.com/nextcloud/nextcloud.com/search?utf8=%E2%9C%93&q=%22support%40nextcloud.com%22 ;-)

jospoortvliet:
@karlitschek what address should we use as general contact address?

karlitschek:
Let's not have an email address here. This just attracts spam and not suitable messages. The contact form should be enough.

MariusBluem:
According to german law we must provide 2 ways to contact the provider of the webpage. In general, providers run into problems because they don't want to provide a phone number ... in our case the opposite is the case 😅 We have the phone number, which is great 💪 

Especially §5 TMG (2) 

> Angaben, die eine schnelle elektronische Kontaktaufnahme und unmittelbare Kommunikation mit ihnen ermöglichen, **einschließlich der Adresse der elektronischen Post**,

For that, I think we should provide _at least_ a contact form (or link to the existing one on the Enterprise page). 

If we don't want to use the support@ email here, we may also want to consider about a use of the abuse@ here (...or a renaming of it into legal@ to describe it better).

MariusBluem:
https://github.com/nextcloud/nextcloud.com/commit/750002f6b38baeef35c64e18e7feae9cd2d484be removed the email without replacement (contact form) 😢

MariusBluem:
THX 👍

PR #312, opened 2016-12-14
Replace support@ with contact form on privacy page

MariusBluem:
We have done this already for the imprint ... there is another place where this is still appearing 🙀 
Lets remove it there too 😉 

@jospoortvliet @LukasReschke 

cc @karlitschek for legal aspects


MariusBluem:
See https://github.com/nextcloud/nextcloud.com/issues/271

PR #323, opened 2017-01-02
Update page-homepage.php with new nextcloud version

mat-mo:


MariusBluem:
Haha ... Good catch 😅

MariusBluem:
We also need to update the text. I will do this in another PR :)

jospoortvliet:
@mat-mo - GREAT catch. Let's get this live ASAP hehe

PR #326, opened 2017-01-02
Footer: Space between RSS and Feed

MariusBluem:
It should be **RSS Feed** instead of **RSSFeed** 😉 

@jospoortvliet @Espina2


jospoortvliet:
Thanks!!! Love the attention to details.

PR #342, opened 2017-01-16
[RFC] Improvements enterprise

Espina2:
@jospoortvliet 

Test and merge if it works

- [x] Should hide and show the Header when we have less than 600px height.

- [x] Should adjust submenu to the top if exist

- [x] Slideshow on Mobile have the same style as the desktop

- [x] Removed arrows on mobile slideshow

Need to look to a better image


jospoortvliet:
all done, great work

PR #349, opened 2017-01-17
Typography fixes

jancborchardt:
- fix typography squishing through tracking reduction (letter-spacing: -1px and -2px), it just made the letters look pressed together
- fix line-height of headings to prevent ascender/descender touching, see below:
Before:
capture du 2017-01-17 13-52-25
After:
capture du 2017-01-17 13-53-39

Please review @nextcloud/designers @nextcloud/website :)


assets/css/custom.scss
```diff
@@ -332,7 +330,7 @@ li {
 		height: 70px;
 		border-bottom: solid 1px #e4e4e4;
 		border-top: solid 1px #e4e4e4;
-		text-transform: uppercase;
+		text-transform: capitalize;
```
jancborchardt: I would prefer if we just use the proper text in the HTML and not do any messing around with text-transform. There’s no reason to do this with CSS. :)

PR #363, opened 2017-01-18
Fix release date

nickvergessen:
@jospoortvliet found it while reading: https://philippe.scoffoni.net/comment-choisir-owncloud-nextcloud/


jospoortvliet:
good catch ;-)

PR #374, opened 2017-01-27
Update page-events.php - first version

cyberanca:
First version of the events page. Added meetups & upcoming conferences


jospoortvliet:
Thanks a lot - I forgot, so I sent the newsletter and @LukasReschke had to tell me, but it is all live and good now 🗡 
we can improve the images after FOSDEM!

PR #376, opened 2017-01-28
Add events to navbar and footer

MariusBluem:
Follow up of #374 😉 @jospoortvliet @cyberanca 


Espina2:
Before that I think It would be nice if the page can look like the new pages. 

Its that possible?

jospoortvliet:
@Espina2 sure, I would not mind - we can do that during this week, perhaps, or next week...

After FOSDEM we should have some nice pictures to use ;-)

jospoortvliet:
So it is after FOSDEM - working on those pictures ;-)

Sadly we don't have a good, high res crew/booth pic, I should've taken that but forgot :sob:

jospoortvliet:
I've updated the Events page to our new style. It is live - review/comments welcome.

https://github.com/nextcloud/nextcloud.com/commit/2bfa4c5e20898aab5121ef6f443b703d0742cd8b

jospoortvliet:
so let's merge now, the events page is there and is awesome ;-)

PR #380, opened 2017-01-30
NC11 on the front page

### LINKED ISSUE #378: Mentions of NC10 on front-page
Now that NC11 was released, shouldn't we update the page to talk about NC11 rather than 10?

jospoortvliet:
fixes #378


jospoortvliet:
works here so... let's put this live.

PR #381, opened 2017-01-30
add providers and optim img

jospoortvliet:
Add half a dozen new providers and optimize some images


jospoortvliet:
working locally so - let's go live.

PR #388, opened 2017-02-03
Remove Release Announcement

MariusBluem:
...otherwise, nobody will look at it, when we release the next version - or have some amazing news 😅 
@jospoortvliet 

### DIFF
--- a/page-homepage.php
+++ b/page-homepage.php
@@ -11,10 +11,10 @@
 	<div class="container">
 		<div class="row">
 			<div class="col-md-6 toptext ">
-				<a class="announcement" href="https://nextcloud.com/blog/nextcloud-11-sets-new-standard-for-security-and-scalability/">
+				<!--<a class="announcement" href="#">
 				<span class="type">Fresh</span>
 				<span class="message"><strong>Nextcloud 11 has arrived.</strong> Take a look at the new features</span>
-				</a>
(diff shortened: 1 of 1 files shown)


jospoortvliet:
we should wait until we have some other amazing news ;-)

MariusBluem:
I think that when it is there always, the users do not longer look at it, since it does not stick out of the "normal" behaviour of the Homepage

jospoortvliet:
Well, it's mostly for new users anyway - I doubt many existing users visit that page for anything other than to find the download page. I think we should always have something there and atm there's nothing really better.

MariusBluem:
What do you think @jancborchardt ... do you also think we should keep this release announcement all the time 😕

jospoortvliet:
Maybe when we open the call for papers for the conference we can put that there?

MariusBluem:
Yap... Nice Idea... But my problem is, that this badge is meant as eye-catcher which should only be used for 2 weeks (more-or-less) if we have important news... normally we should only have this "A safe home for all your data" stuff there - otherwise - as I've said, it will loose its "eyecatcher-function" - but you are our marketing hero 🙈

jospoortvliet:
so how about we close this and link to our security scanner? ;-) will make a PR.

PR #392, opened 2017-02-05
Outlook: Use GitHub instead of Sourceforge

MariusBluem:
@jospoortvliet 

### DIFF
--- a/page-outlook.php
+++ b/page-outlook.php
@@ -3,13 +3,13 @@
 	require(["require.config"], function() {
 		require(["pages/outlook"])
 	});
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Yeah... I linked to Sourceforge because it is kind of their 'homepage' at the moment.

But this is OK, ppl can get it right away and I honestly am not a fan of Sourceforge anymore either, so... ;-)

Plus, I talked to the developer at FOSDEM, he's working on a better site etc anyhow so this is OK until he has that stuff done.

PR #393, opened 2017-02-05
Use subdomain for example Federation ID

MariusBluem:
...the most Instances use `cloud.` as subdomain 😅 More a cosmetic papercut - but yes 😁 
@jospoortvliet 


jospoortvliet:
Without your papercuts this site would quickly deteriorate. It is starting to look like "Jos makes a mess. Marius cleans up"...

PR #398, opened 2017-02-08
start adding pilling.js in the slider

jospoortvliet:
Not sure if this will go anywhere or should be removed?


Espina2:
Yes remove this. Its a test I made a time ago.

PR #412, opened 2017-02-21
trying to fix mobile bg making clicks in header impossible

jospoortvliet:


jospoortvliet:
@Espina2 so this seems to work quite OK, could you test it too? If it works for you I guess we should merge  ;-)

Espina2:
This breaks the animation. I should look at this.

jospoortvliet:
going to staging for more testing...

PR #413, opened 2017-02-21
make the Migration page have a few more quotes on top

jospoortvliet:
* [ ] check if everyone is good being quoted


jospoortvliet:
tested on staging, works.

Issue #415, opened 2017-02-22
Content initially shown directly, then flashed away, then faded in again


jancborchardt:
This happens on every page and looks strange:
https://cloud.jancborchardt.net/s/ILN4zJnH8upREwV

cc @Espina2 @nextcloud/website @nextcloud/designers

Espina2:
Its probable due to the original state of the document doesn't have opacity 0 so when it load the javascript file it adds animate from opacity 0 to 1. I will check this when I have free time.

jospoortvliet:
It still happens on the Enterprise page somehow.

PR #417, opened 2017-02-22
[RFC] fix flickering as mention in #415

Espina2:
This can be merge

https://github.com/nextcloud/nextcloud.com/issues/415

Pls review @jancborchardt @jospoortvliet


jospoortvliet:
Not sure where my reply went but - it works for me except on the enterprise page. That one still has flickering when each item appears...

jospoortvliet:
It also still happens on the 'verified security' paragraph on the secure page.

PR #430, opened 2017-03-17
link to security scan

jospoortvliet:
How about not removing like proposed in #388 but replacing with this?

### DIFF
--- a/page-homepage.php
+++ b/page-homepage.php
@@ -11,9 +11,9 @@
 	<div class="container">
 		<div class="row">
 			<div class="col-md-6 toptext ">
-				<a class="announcement" href="https://nextcloud.com/blog/nextcloud-11-sets-new-standard-for-security-and-scalability/">
+				<a class="announcement" href="https://nextcloud.com/blog/nextcloud-releases-security-scanner-to-help-protect-private-clouds/">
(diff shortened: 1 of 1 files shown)


jospoortvliet:
@MariusBluem because you care and @NinaNC because you should 😜

page-homepage.php
```diff
@@ -11,9 +11,9 @@
 	<div class="container">
 		<div class="row">
 			<div class="col-md-6 toptext ">
-				<a class="announcement" href="https://nextcloud.com/blog/nextcloud-11-sets-new-standard-for-security-and-scalability/">
+				<a class="announcement" href="https://nextcloud.com/blog/nextcloud-releases-security-scanner-to-help-protect-private-clouds/">
 				<span class="type">Fresh</span>
-				<span class="message"><strong>Nextcloud 11 has arrived.</strong> Take a look at the new features</span>
+				<span class="message"><strong>Nextcloud releases Security Scanner.</strong> Check if your data is secure!</span>
```
MariusBluem: "has released"

jospoortvliet: I made it 'released a' so the past tense is there but it doesn't get too long ;-)

PR #431, opened 2017-03-17
add Netways to providers

jospoortvliet:


jospoortvliet:
@NinaNC @MariusBluem any chance you can have a quick look and yay/nay?

PR #432, opened 2017-03-20
[Rework] Devices

MariusBluem:
unknown-5

@jospoortvliet @karlitschek


Espina2:
On the spreedbox I think it works better with the image on the right and the text on the left.

Espina2:
I didn't like the header image, can we find another one?

MariusBluem:
> I didn't like the header image, can we find another one?

Have you something specific in your mind? 😄  

> On the spreedbox I think it works better with the image on the right and the text on the left.

Yes, but how to do this on mobile 😕

Espina2:
> Have you something specific in your mind? 😄
Not yet, but I can try looking for something later.

> Yes, but how to do this on mobile 😕

block {
float: left;
}

@media (max-width $break-small) {
block {
float: none;
}
}

On mobile the image should be on top of the text not on the side.

Hope this helps. :)

jospoortvliet:
I absolutely LOVE the header image but hey, that's just me :dagger: 

Yeah it is probably good to alternate the images left/right. We should put the bloc float right css in custom.css so it can be used everywhere btw.

jospoortvliet:
let me do that

Espina2:
The images should be more human, the image is okay. But is to concrete and close up. In general, the images work better when having a person on it. 

It's very difficult to choose good pictures, so I never complain too much about that. But in the end, they should look that all are part of the same "family". 

For example this image on the transifex website, it looks good because its not literal and have exactly the things I mention. 

https://www.transifex.com/ 

Hope this help to choose better pictures in the feature. Maybe I should write this somewhere. :)

jancborchardt:
Great page relayout! :)

I’d keep with images on the left. This is a product page and you want to directly compare. The right-left is a bit too playful here.

Agree with @Espina2 about the header image, but it’s very hard to find a fitting picture. Just spent some time searching both [Pexels](https://www.pexels.com/) and [Pixabay](https://pixabay.com/) but no success.

MariusBluem:
> Agree with @Espina2 about the header image, but it’s very hard to find a fitting picture. Just spent some time searching both Pexels and Pixabay but no success.

I can offer those two from free sources, but I am also not happy with them. No people. I thought the picture I've taken in my PR is the best of this three 😁 

hdd-75210
technology-1587673_1920

> I’d keep with images on the left. This is a product page and you want to directly compare. The right-left is a bit too playful here.

After testing it, I agree with you. Lets keep it as it is. We just need to make the Spreedbox a bit smaller 😁 Syncloud working fine.

Espina2:
@MariusBluem "bring you server home", why not a people at home? I photoshop the device to be for example on a table. 

> I’d keep with images on the left. This is a product page and you want to directly compare. The right-left is a bit too playful here.
> After testing it, I agree with you. Lets keep it as it is. We just need to make the Spreedbox a bit smaller 😁 Syncloud working fine.

The logic in all the parts of the website is how I said, and makes sense to use the same logic here to. 
Just want to point that.

Espina2:
example of better picture
chirp_accessories-module_1440_4

jospoortvliet:
So, problem solved, we have a new picture :laughing: 
10502208-0fbe-11e7-9980-770f4f3b22f5

Espina2:
your skills at gimp!!! 🥇

jancborchardt:
@Espina2 yeah, a gimped Nextcloud box on a table would be great!

jospoortvliet:
@jancborchardt well, I did that :rofl:

jospoortvliet:
How about https://cloud.nextcloud.com/s/pyateHxgXOzsXC8 ... ?
or https://cloud.nextcloud.com/s/ECXBBOOoyFunitZ ;-)

Espina2:
@jospoortvliet 

You are mixing images with different perspectives, it will be very difficult to not look photoshopped.

Try something simpler like this, you can have a random device each time the page is loaded. 

hero_example

Espina2:
Edited photos here: 

https://cloud.paulomoura.com.pt/owncloud/index.php/s/9XSA78XmXxm8xgT

jospoortvliet:
So I prefer to have this page updated with the image we have in it now, rather than not at all. So unless there's an easy to use image from pexels or so we can drop in, let's just merge this as is... getting stuck forever because we want perfection instead of clear improvement is pointless.

jancborchardt:
I agree, let’s please move forward. I think the intial screenshots @MariusBluem posted above look great so let’s go. :+1:

jospoortvliet:
Tested, deemed awesome. merging!

PR #436, opened 2017-03-22
No more twitter for rullzer

rullzer:
It turns out my mind is not organized enough to successfully communicate in <=140 characters.

Also I hate that blue bird.


jospoortvliet:
lol

PR #439, opened 2017-03-31
crazy rough mockup of new install page

jospoortvliet:
So I tried to implement a mockup that Jan made - this might or might not be very useful as a start ;-)

@Espina2 would this be helpful or do you prefer to start from scratch?

install nextcloud 3-30-2017 3-03-19 pm


Espina2:
I think I will use the other PR because it already have many things that we need. But let this be here, so I can see the information you added.

PR #443, opened 2017-04-03
New External Storage page

jospoortvliet:
Not SUPER good, but a decent start I think:
storage nextcloud 3-4-2017 21-06-06
@Espina2 @NinaNC would love to hear feedback on style and content.


NinaNC:
My two cents :)
816983f2-18b1-11e7-90e4-26fb43ebf314

Espina2:
@jospoortvliet sounds good to me, just one tip on the Public, Private and hybrid unified the technologies you can put in one or two lines instead? 

Try to maintain the same space used for each item, and it will look better.

MariusBluem:
Could we also show some logos of the external storages instead of this list? 
This would reduce space-wasting on left and right side 😁 @jospoortvliet

templates/header-top-navbar.php
```diff
@@ -39,6 +39,7 @@
 						<li class="nav__item"><a href="<?php echo home_url('sharing') ?>"><?php echo $l->t('Sharing');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('clients') ?>"><?php echo $l->t('Clients');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('workflow') ?>"><?php echo $l->t('Workflow');?></a></li>
+						<li class="nav__item"><a href="<?php echo home_url('storage') ?>"><?php echo $l->t('Storage');?></a></li>
```
MariusBluem: I think "External Storage" or something like "Storage integration" (for somebody who is not **that** technical) would describe this page better 😁

jospoortvliet: I'd rather stick with one word as the others do... Already thinking about cutting 'Outlook integration' to 'Outlook'... But we should feature Outlook integration on nextcloud.com/features first so ppl find it easier.

jospoortvliet:
Here is the modified version, let me know what you think - it deals better with the issue @MariusBluem pointed out ;-)
storage nextcloud 4-4-2017 18-42-28

MariusBluem:
Some feedback:
bd93e560-1966-11e7-8cff-e282e273727b

jospoortvliet:
Fixed headline and added a call to action... ;-)
storage nextcloud 4-4-2017 19-40-35

Espina2:
It looks nice. Just one question, why you have that blue claim before the last point? 
@jospoortvliet

PR #459, opened 2017-04-28
updated the video

NinaNC:


NinaNC:
@MariusBluem I honestly don't know but you're right, fixing that

PR #468, opened 2017-05-19
Update information about NC 12 migration

nickvergessen:


page-migration.php
```diff
@@ -109,6 +110,7 @@
 		<li class="section--paragraph"><?php echo $l->t('File Firewall');?></li>
 		<li class="section--paragraph"><?php echo $l->t('File Automated Tagging');?></li>
 		<li class="section--paragraph"><?php echo $l->t('Files Drop');?></li>
+		<li class="section--paragraph"><?php echo $l->t('SharePoint');?></li>
```
nickvergessen: Not sure how to best mention customgroups/circles here, they are migrated, but they should still be checked if everything went fine.

jospoortvliet: I'd always check after a migration - so this is no different from other parts.

PR #478, opened 2017-05-27
We call it File Drop

### LINKED ISSUE #469: Secure Drop -> File Drop
According to https://github.com/nextcloud/server/pull/4943, please update the page, or (I would prefer) merge it with sharing 😉 

@jancborchardt @jospoortvliet

MariusBluem:
Fixes #469 @jospoortvliet @NinaNC 

- [ ] should we rename /page-secure-drop.php to /page-file-drop.php - better now, than later? 🤔 


MariusBluem:
Made a mistake. Unfortunately this is now part of #479 🙄

PR #480, opened 2017-05-29
Update page-compare.php

jospoortvliet:
describe us vs them ;-)
As suggested by Tom: https://twitter.com/tombrossman/status/869160369201176576

what do you think @karlitschek ?


jospoortvliet:
OK'ed by the Frank :whale2:

PR #487, opened 2017-06-10
SCSS information in readme

ViRPo:
- add SCSS and css map notes to readme file


ViRPo:
@jospoortvliet this is the readme change we talked about
do you think "Notes" section is a good place for this type of instructions?

jospoortvliet:
Yeah, I think it'll work. Thanks a lot!

PR #495, opened 2017-06-13
Fix SCSS readme note

ViRPo:
As it turns out, I was wrong with my notes aboud compiling SCSS
Truth is, every CSS file currently has sourceMappingURL set at the end of the file, which means, that all SCSS files were originally compiled with sourcemap option turned on.
Not compiling them together with sourcemap would mean deleting sourceMappingURL from the css file, which is unwanted

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,5 @@ process easy and effective for everyone involved.
 ### Notes
 
 * Please don't commit straight into the master or live branches, these branches should remain as stable as possible, and changes should be discussed amongst the community.
-* After changing any SCSS file, it should be recompiled to CSS before merging. Following options are recommended:
-  * Unix: `sass file.scss:file.css --sourcemap`
-  * Windows: `node-sass file.scss file.css --source-map true`
+* If you are chaning SCSS files during development, following command for automatic compilation is recommended:
+  * `sass --watch [location to scss files]`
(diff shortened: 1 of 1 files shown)


jospoortvliet:
mja, I just use `sass --watch *.scss`
and `sass bla.scss bla.css` and never got in trouble but this might be more correct ;-)

Espina2:
We just need to run sass --watch path to assets.

ViRPo:
I see, what would you suggest to write to readme note?

jospoortvliet:
Maybe just add to what you wrote:
During development it is useful to use the `--watch` option, running the command: `sass --watch [location to scss files]` which will instantly update the CSS files whenever you modify the scss ones.

ViRPo:
It's a small step for a man and small step for a humanity, but I finally figured out, what the recommended settings should say
There are many recommended ways to achieve that in Windows, but `sass --watch [location]` should work just fine for most cases even in Windows :)

Sorry for the mess

jospoortvliet:
:+1:

jospoortvliet:
Thanks a lot! Every little bit helps - I know @NinaNC is setting up a website dev environment so this will certainly help her :D

NinaNC:
Current state: 


alt="I'd better do this during the week-end"

PR #434, opened 2017-03-21
[WIP] New install Page

Espina2:
A brand new look install page.

At the moment almost the content is "dummy" and are only made the server part.

@LukasReschke don't worry I will update the downloads links with the strings defined.

@nextcloud/designers what do you think?

install page

PR #500, opened 2017-06-17
Improve installation

MariusBluem:
- 7f04ee5 will remove the possibility to download Nextcloud 12 RC3 over the release-channels page
- 7835a4e was discussed in the forum already. Buying a Nextcloud Box or finding a provider has nothing to do with download, so lets go for "Get Started"
- bb2d666 links the changelog to release channels, and adds a missing `stable12`-anker. Small detail fixes.
- 3c66001 are small style changes (moving newsletter link to right side, so that it is always visible. Link to release channels.

@jospoortvliet @NinaNC


jospoortvliet:
They are generally fine. The removal of the download link has no effect as the download doesn't show up when the version number above it is empty, but it is fine ;-)

The download to Get Started is something I would want to test.

We once upon a time changed it from 'install' to 'download'. The reason was that people who are looking to get Nextcloud are looking for the word 'download' on our website. If you change that into another word, it simply becomes harder to find Nextcloud. The marketing team told me there's plenty of research that shows a change of words away from a super familiar term ('download') really lowers the number of people downloading your product... And we don't want that.

So I'm open to this change, but ONLY after we have set up some way of tracking what percentage of people clicks there and downloads it. Some kind of mouse heatmap thing has to be in place first so we can see the effect and, after a few weeks, decide to keep it or undo it.

Right now we might be hurting our growth without even knowing it.

Am I crazy to be so worried?

MariusBluem:
> The removal of the download link has no effect as the download doesn't show up when the version number above it is empty, but it is fine ;-)

I thought that too, but then I saw: https://nextcloud.com/install/#testing-development

Espina2:
Don't forget about that this PR that will be complete very soon, https://github.com/nextcloud/nextcloud.com/pull/434 . Tomorrow I will 99% that is finished, so see if it makes sense to change to get started. 

I think "Get started" remind me a page that is more descriptive and help you step by step.

Espina2:
@jospoortvliet You can use https://www.hotjar.com/ I can help you with that, I am very familiar with this type of stuff. If you want help just call me in one of this days. I have with more time this days. :)

MariusBluem:
> Don't forget about that this PR that will be complete very soon, #434 .

Nice 🙈

> Am I crazy to be so worried?

Just a bit. 😜 I think the barrier to get/install Nextcloud should be as low as possible. An Installation Page (install) reminds me of handling PHP, Webserver and such stuff, which isn’t wrong at all ... but the beginners or not apache-natives should know, that there are alternatives which can be found on this side too. With alternatives, I mean Devices and Providers. Maybe also the snap or docker-image, but I am not a big fan of giving images in the hand of people who don’t know what they do ...

> I think "Get started" remind me a page that is more descriptive and help you step by step.

Another alternative would be **Get it** (seen on openSUSE) ...

jospoortvliet:
@Espina2 I'd prefer to use https://plugins.piwik.org/ClickHeat as we already have piwik... Let's see.

Anyhow, @MariusBluem I would like to keep out the change to the top bar for now - Download will do the trick for now and we can try other things once we have tracking up.

The rest - let's merge it, while we finish the install page... the work is done anyway.

Espina2:
Sounds like a plan for me. I didn't know that he are already using something so its okay for me. :)

MariusBluem:
> The rest - let's merge it, while we finish the install page... the work is done anyway.

👍 Please merge!

PR #501, opened 2017-06-17
[RFC] fix image issues on homepage

Espina2:
Fix issues with image on homepage.


jospoortvliet:
I think it would be good to make the image wider, I already have some empty space on the right on a full screen window and it could be worse on wider windows.

Espina2:
You still see that with this commit? @jospoortvliet

jospoortvliet:
yes

Espina2:
2560 widther screenshot above, can you show me what happens in your. We should fix that, and it should be easy. :)

screen shot 2017-06-18 at 17 58 00

jospoortvliet:
Ha, it might have been a caching issue as it doesn't happen anymore. So - solved?

PR #509, opened 2017-07-06
added the free as in beer version

NinaNC:
still waiting for the download link - don't merge please

EDIT: ready to review/merge.


MariusBluem:
Maybe we should not call it „Free as in beer“-Version on a Features-page 🤔

MariusBluem:
@NinaNC Any news on this?

NinaNC:
Sorry, missed your comment. Yes, you're right. I want to wait for the update about the paid version to fix everything at once.

page-outlook.php
```diff
@@ -77,6 +77,24 @@
...
+<section class="file-sharing">
+	<div class="container">
+		<div class="featurerow">
+<!-- 			<h2 class="section--heading-1 section--text--center"><?php echo $l->t('Free-as-in-beer version');?></h1> -->
+			<div class="row">
+				<div class="col-md-6 floated revealOnScroll">
+					<a href="<?php bloginfo('template_directory'); ?>/assets/img/outlook/fileupload_3.png"><img class="img-responsive" src="<?php echo get_template_directory_uri(); ?>/assets/img/outlook/fileupload_3.png" alt="in action" /></a>
+				</div>
+				<div class="col-md-6 featureblock revealOnScroll">
+					<h2 class="section--heading-2"><?php echo $l->t('Free-as-in-beer version');?></h2>
```
jospoortvliet: Put it as 'free for home users', that should be fine.

page-outlook.php
```diff
@@ -77,6 +77,24 @@
...
+					<p class="section--paragraph"><?php echo $l->t('Nextcloud strives to provide private cloud solution for everyone so we worked with our partner to provide a <a href="https://download.nextcloud.com/outlook/20170706/">free version of the Secure Outlook Add-in for home</a>. Note that the free version is only meant for home users and small organizations.');?></p>
```
jospoortvliet: A date in the link seems not very... easy. Let's instead create a variable and use that. See the 'strings.php' for how that works and how it is used on the install page. it's some <php bla> code, not super hard ;-)

We shouldn't link to the download page but put the 2 or 4 different versions as buttons on this page, I think.

page-outlook.php
```diff
@@ -77,6 +77,24 @@
...
+					<p class="section--paragraph"><?php echo $l->t('The free Secure Outlook Add-in features the core capabilities: sharing from your Nextcloud instance instead of sending insecure large attachments, with an optional expiration date. Branding is limited, and it only works for less than 50 users detected on the system.');?></p>
```
jospoortvliet: remove the 'less than 50 users' bla. Sure, we put in limitations. Just note it is designed for home users and small businesses. What that means exactly doesn't matter, we might change the limits, remove them, put in others etc depending on how the add-in is used. If we find a company with 5K users using it, we clearly have to change the rules... We should also consider putting in a license that doesn't allow usage in organizations > 50 employees or something. Let's see.

jospoortvliet:
See my inline comments. Also, we have to update the screenshots ;-)

jospoortvliet:
Made the changes I requested and put in screenshots. While one screenshot is missing, it's quite an improvement already and I propose to put it live and do another iteration once a new version of the free version is out or when I get the screenshot in or so...
outlook nextcloud 3-8-2017 22-45-03

jospoortvliet:
@MariusBluem could you have a look at the screenshot and tell me what you think? @Espina2 your input is also welcome!

jospoortvliet:
reviewed by Marius - and OK ;-)

PR #515, opened 2017-07-12
[NI/WIP] Test browser features and added a way to a browser without javascript see the website

Espina2:
@jospoortvliet 

I made the first step to have a website that works with no javascript. Off course, not everything it will work, but 90% it's fixable. The idea is very simple, I make a verification for seeing if javascript is enabled, if not a no-js class it will be at the html tag. So you can use something like the code above to override the css properties you need. I suggest that you made something like this in the scss page file.
```
.no-js body {
/// CODE HERE
} 
```


jospoortvliet:
Tested it on a dozen pages. Working, well, quite acceptable. Of course reduced functionality but the menu works, I fixed the header images... It is quite good. You can even download Nextcloud itself, though not the VM's and such ;-)

Espina2:
Yup, I think that we should have a warning at top of the page when javascript is disabled, something like this: "We love privacy to, but in order to have a better experience you should enable javascript" 

@jospoortvliet what you think?

jospoortvliet:
Yeah, that's a good idea. But that would mean changing all pages, I guess, and adding javascript that removes the warning? Or can it be done in the common php?

Espina2:
You show the warning by default and if javascript is enabled, it removes the warning. :)

jospoortvliet:
Totally sensible. Header, I would say. Something like:
```
<div class="nojs-warning">
<div class="container text-center">
<p>You have javascript disabled. We tried to make sure the basics of our website work but some functionality will be missing.</p>
</div>
</div>
```
I could add it to the header and make it look nice, could you try to add the JS needed to hide it on normal cases?

Espina2:
Don't merge it, let me check one thing first. I'm not in front of the computer.

Espina2:
I think it's better to add it on the footer, or in the index, so it parses the first part of the document and do the tests, if js is enabled at that point it removes.

jospoortvliet:
It'd be nice if it was quite visible, though. The footer - nobody sees that 90% of the time...

Espina2:
@jospoortvliet only the markup it will be on the footer. The ideia is to have an absolute diverse on the top of the document. I only want to do that for not showing at all when js is enabled. :) the way you have it, it will show and just after a while it will go away.

jospoortvliet:
@Espina2 ah! Forgive my lack of understanding... I will put it in the footer.

Espina2:
@jospoortvliet no problem im not complete sure that it will fix the problem, its because of that, that I will take a look. :)

Espina2:
@jospoortvliet @MariusBluem please review. I added a warning when we have the js disabled, and and the dropdowns menus a white background.

jospoortvliet:
tested on 15 different pages with and without javascript. Doesn't break anything, merging.

PR #518, opened 2017-07-14
Enforce linting

LukasReschke:
Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,26 @@
+pipeline:
+  clone:
+    image: plugins/git
+    depth: 1
+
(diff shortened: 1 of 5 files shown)


Espina2:
@LukasReschke its possible to do the same with sass and javascript?

LukasReschke:
> @LukasReschke its possible to do the same with sass and javascript?

Let me check :)

Espina2:
Thanks if so, let me know wich linters you install so I install the same on the editor. ;)

PR #520, opened 2017-07-15
Conference on front

jospoortvliet:
Put the conference on the front page & link to the Friday Focus blog on the conf page.


jospoortvliet:
@MariusBluem @NinaNC if you could have a look and spot any big issues... would rock. If OK, update branch & merge pls ;-)

page-homepage.php
```diff
@@ -12,9 +12,9 @@
 	<div class="container">
 		<div class="row">
 			<div class="col-md-6 toptext ">
-				<a class="announcement" href="https://nextcloud.com/blog/welcome-to-nextcloud-12/">
+				<a class="announcement" href="https://nextcloud.com/conf/">
 				<span class="type">Fresh</span>
```
MariusBluem: Call it Event instead of Fresh and change the color to (I would say red) please 🙈

Espina2: @MariusBluem why red? Orange is complementary color of blue so let it orange. Don't use no sense colors if you have no explanation for that.

Espina2: Anyway I agree with the text

jospoortvliet:
OK?

MariusBluem:
:+1: Content wise 😏

PR #522, opened 2017-07-15
fix secure drop page

jospoortvliet:
the secure-drop page has broken JS due to the yt plugin. This should fix it! pls check, @Espina2 !!


Espina2:
@jospoortvliet this should work as expected.

PR #527, opened 2017-07-19
add case study

jospoortvliet:
pls review @MariusBluem ;-)

PR #529, opened 2017-07-19
Make website less insecure…

LukasReschke:
Ref https://github.com/nextcloud/nextcloud.com/pull/527#issuecomment-316327040

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/page-architecture-whitepapersubmit.php
+++ b/page-architecture-whitepapersubmit.php
@@ -93,7 +93,7 @@ if(isset($_POST['email'])) {
 <section class="section--whitepaper">
     <div class="container text-center">
         <h3>Thank you for your interest in our whitepaper!</h3>
-        <p>The whitepaper has been sent to <?php echo $email_to ?>,</p>
+        <p>The whitepaper has been sent to <?php echo htmlentities($email_to) ?>,</p>
         <p>check your spam folder if you can not find it!</p>
     </div>
 </section>
--- a/page-auditing-whitepapersubmit.php
+++ b/page-auditing-whitepapersubmit.php
@@ -80,7 +80,7 @@ if(isset($_POST['email'])) {
 <section class="section--whitepaper">
     <div class="container text-center">
         <h3>Thank you for your interest in our datasheet</h3>
-        <p>The datasheet has been sent to <?php echo $email_to ?>,</p>
+        <p>The datasheet has been sent to <?php echo htmlentities($email_to) ?>,</p>
         <p>check your spam folder if you can not find it!</p>
     </div>
 </section>
--- a/page-file-access-whitepapersubmit.php
+++ b/page-file-access-whitepapersubmit.php
@@ -80,7 +80,7 @@ if(isset($_POST['email'])) {
 <section class="section--whitepaper">
     <div class="container text-center">
         <h3>Thank you for your interest in our datasheet!</h3>
-        <p>The datasheet has been sent to <?php echo $email_to ?>,</p>
+        <p>The datasheet has been sent to <?php echo htmlentities($email_to) ?>,</p>
         <p>check your spam folder if you can not find it!</p>
     </div>
 </section>
--- a/page-gs-whitepapersubmit.php
+++ b/page-gs-whitepapersubmit.php
@@ -80,7 +80,7 @@ if(isset($_POST['email'])) {
 <section class="section--whitepaper">
     <div class="container text-center">
         <h3>Thank you for your interest in our whitepaper</h3>
-        <p>The whitepaper has been sent to <?php echo $email_to ?>,</p>
+        <p>The whitepaper has been sent to <?php echo htmlentities($email_to) ?>,</p>
         <p>check your spam folder if you can not find it!</p>
     </div>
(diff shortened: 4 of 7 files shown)


Espina2:
@LukasReschke cool branch name. ehhe

jospoortvliet:
so now we have checks which take half an hour to an hour to complete. Great but that doesn't work with time sensitive stuff :-1:

LukasReschke:
> BTW I copy-paste the form you wrote ;-)
> we should do this with mautic of course but that's not done yet...

That is not true. The $email_to has been added by you. The original form does *NOT* contain this.

> so now we have checks which take half an hour to an hour to complete. Great but that doesn't work with time sensitive stuff 👎

No. The checks take seconds talk to the big master to get some more CI servers faster. This would help. Sorry but there's no way we're continuing this "YOLO deployments". Do your PRs earlier, wait a bit or get some more CI servers. In the server we also have to wait so long.

Nothing is as time intensive that it can't wait 30 minutes. If you need to do a lot of patches later on this just means it hasn't been tested good enough before and shouldn't be deployed in the first place.

MariusBluem:
@LukasReschke can push it, and maybe (if needed) set all caches to Invalid, so that every user (and not only new visitors) get this new and fixed page.

jospoortvliet:
The problem is that timing is important for this. Stuff needs to be pushed, tested and published in under 30 minutes - I don't want to put this out before 11 and it has to be on the website at 11:30.

Many things (including full releases) can't be on github before 11 because we're working on marketing things here we try to keep under wraps.

And they have to be out by 11:30 because we want the maximum impact and that is the optimal time.

And yeah, small typo's are inevitable unless we put our website entirely private and work on it there. So we need a quick turn-around and not checks which still say 'build is pending' 35 mins after something has been submitted.

MariusBluem:
> And they have to be out by 11:30 because we want the maximum impact and that is the optimal time.

And it has to be 11:30 **today**. See no problems for the future. Open PRs one day before -> reviewing -> CI -> merge -> test -> deploy when needed 😇

LukasReschke:
I see the need for this but the fact is that we sadly only have 1 CI infrastructure which consists of 3 servers which are all totally overloaded and thus in working times the slots are simply already all used.

And no, having no CI at all is not a solution. Sure "typos and so on" may be a valid reason. But in this case just create a new branch from master, push there and deploy this one by doing a `git checkout $branchname`. 

(which is still SUPER far from ideal and I'd still vote for just doing PRs earlier or moving to a proper CMS where we don't have the issue of: "a single typo fix can break the whole website")

rullzer:
Things needs to be tested. There is no way around that. CI is there exactly for that reason.

If stuff is so incredible time sensitive it can't be put up earlier. Then maybe move this to a private repo?

We have had more than 1 occasion where a faulty deploy resulted in a broken website. Which is just not acceptable.

jospoortvliet:
If I check out a branch I break the entire website. Been there, done that, got yelled at - quite a few times. If you can make that work, sure, fine with me.

> And it has to be 11:30 today. See no problems for the future. Open PRs one day before -> reviewing -> CI -> merge -> test -> deploy when needed :innocent:

No, this constraint is relevant fo every major thing we do: we keep it hidden until 30 minutes before the blog post is published and promoted. So we have 30 minutes to review, test and get it live to get the maximum impact. This is for PR and competitive reasons, and is not negotiable: processes will have to adjust to reality, not the other way around.

LukasReschke:
> No, this constraint is relevant fo every major thing we do: we keep it hidden until 30 minutes before the blog post is published and promoted. So we have 30 minutes to review, test and get it live to get the maximum impact. This is for PR and competitive reasons, and is not negotiable: processes will have to adjust to reality, not the other way around.

It is also not negotiable that we need CI. And this will not go away, you can have quicker CI by throwing money at the problem.

MariusBluem:
If things need to be hidden, do this (or maybe parts of the website if possible) in a private repo.

jospoortvliet:
> It is also not negotiable that we need CI. And this will not go away, you can have quicker CI by throwing money at the problem.

Fine, do it, ask for it, mail frank, whatever. Until then, we can't stop our PR and marketing until that CI shit is fixed.

jospoortvliet:
> If things need to be hidden, do this (or maybe parts of the website if possible) in a private repo.

That is what I said - we can alternatively move our entire website private. But that means no more community contributions including @Espina2 which isn't really an option.

MariusBluem:
I think this is the time where we should stop this discussion, and continue it in a calm and constructive way by person!

Espina2:
my two cents. I indeed agree with changing this CMS, on having this repo private, and also in having more concrete pipeline process. PR -> CI -> STAGING -> DEPLOY 

I really think that WordPress should be removed at all cost, it difficult everything, right now we are using to routeing (awesome).  

In the others hands @jospoortvliet are right to, sometimes we need to pull things faster, and sometimes is difficult to plan ahead of time. 

I know that this enterprise stuff but I just want to give my opinion.

PR #533, opened 2017-07-23
typos on contributors-page

MariusBluem:


MariusBluem:
@jospoortvliet Please review!

jospoortvliet:
I keep the dots inside the tags these days because I need to do that with the newsletters or you get weird line breaks ;-)

But it makes no difference, I think. +1 on catching the , :bowing_man:

PR #538, opened 2017-08-03
Pi3 support for the Box

jospoortvliet:
The Raspberry Pi 3 image is now 'official'.


jospoortvliet:
Done trademark & Pi name but we already mentioned the oDroid and for good reason - it is the most powerful hardware and for a more technical user the best solution. That the guides don't use snappy (it could, btw) is not an issue for me. Yes, it is preferred, no, not mandatory... If you install following that guide you have to know what you're doing anyhow...

There's "good" and "perfect" ;-)

jospoortvliet:
Thanks Marius.

PR #542, opened 2017-08-05
add security scanner

jospoortvliet:
relatively minor change...


MariusBluem:
I think we should take the same text as used on scan.nextcloud.com:

_Our scan is strictly based on publicly available information, that is the list of known vulnerabilities relevant for ownCloud/Nextcloud releases as well as any applied hardenings/settings we can scan without having access to the server._

jospoortvliet:
Done ;-)

PR #548, opened 2017-08-08
10.x is not longer supported

MariusBluem:
Just removing the 10.x release from security page, since this is not longer a supported release :)


jospoortvliet:
Yeah, we did its last release. And I thought I should mention that in the blog post but I forgot! Adding it.

PR #531, opened 2017-07-20
[WIP/NI] Better Handling of dependencies

Espina2:
- [x] Remove the timeout from requireJS
- [x]  Use yarn for handling dependencies
- [x] Update jquery version and remove deprecated code
- [x] Use lodash instead of underscore
- [x] Using slim versions of libs instead of importing the whole library
- [x] Organise scss files by page
- [x] Use only the parts of bootstrap needed instead of importing everything
- [x] Remove vendor files from repo
- [x] Ident all the scss files correctly
- [x] Improve animations
- [x]  Inline SVGS for fewer http requests
- [x]  Optimized svgs
- [x]  Code Highlighting

PR #557, opened 2017-08-17
3.1 mb svg to 480kb png

jospoortvliet:

🙈 

The difference is certainly visible if you have a high res screen and switch quickly between tabs - the pattern is more fuzzy.

Also, 480 kb is still idiotically much and we really need some other solution to do this pattern but I have no idea how.

I optimized the SVG down to 1.7 mb without making it look different but that is still too much; a higher resolution PNG was 650 kb which I still thought was too much.

other ideas welcome but I suggest we merge and deploy this first to help the 4K users who visit our home page each day...


jospoortvliet:
Put it on staging: https://nextcloud.com/?staging=true

Espina2:
@jospoortvliet https://github.com/nextcloud/nextcloud.com/pull/531

This is already done here. 

You can do a media queries for smaller svgs or jpeg's on smaller devices.

MorrisJobke:
> #531

@jospoortvliet Could you put this on staging? On @Espina2 server it is already a lot faster according to webpage speed (6s -> 2s) but I guess there is more possible with our tweaked server setup.

jospoortvliet:
What I can tell is that https://nextcloud.com/?staging=true is now completely broken 🗡  so something doesn't work well with our setup...

jospoortvliet:
This might still be useful as a temporary fix. Besides, tinypng made it even smaller... I just keep uploading the files and it keeps shaving off. I see no differences to the originals so I'm good with it if you are...
spectacle hz4474

jospoortvliet:
Deployed, live. Our site should now load significantly faster ;-)

It works here but if you see issues, lemme know.

Espina2:
Good Improvoment. :)

Also we should do that to all our png images. :)

jospoortvliet:
@Espina2 I DID, remember ;-)

But this was a SVG, not a png. And I dunno if you feel like uploading and re-uploading every single image 10x to tinypng.com, but I don't... I decreased the size by 70% already in the earlier PR, the extra 15% (1x) to 30% (multiple times) we can get for all images is nice but not worth my time, sorry.

PR #573, opened 2017-08-31
optimize images

jospoortvliet:
Cut size of png's in half with pngquant & optipng


Espina2:
For png use https://tinypng.com/

It will give the best quality with the smallest file size. Its like magic.

@jancborchardt Already see that magic happen.

jospoortvliet:
@Espina2 tinypng.com uses... pngquant, the tool I used too ;-)

I doubt they can shave of any more space.

jancborchardt:
> I doubt they can shave of any more space.

Just tried it out with some of the first images:
screenshot from 2017-09-06 19-03-53

I’d say avg 14% savings is quite significant on all the images.

jospoortvliet:
Fascinating. Every time you upload the file to panda, it compresses it, you download and upload it again, it again removes some stuff and compresses. Rinse, repeat, it gets smaller each time.

The first time, it compresses just as much as I have done with pngquant and optipng (50-60%), but you can compress multiple times if you re-upload it... It of course reduces the colors each time, using dithering, but it is barely visible on most images. Well, I can run them through the Panda one or two more times, for sure... Each time, it seems to save less of course. 50% first (that is what pngquant already did), then 15%, 10, 5...

We already got a ~50-60% reduction. Another 15% could be nice. We can spend more time and get another 10% and run it a third time and get 5% more...

Question if it is worth the time and effort... Not my time, to be honest. The 60% was worth some effort but uploading and downloading to/from panda for another 30% (uploading & downloading each image 3-4 times) - pfff.

jancborchardt:
Btw before that's done in amy case - we should make sure we have all the svg sources or retain the high quality PNG still. Even if it's just dithering or so, future resolution improvements and such will bite us. ;)

jospoortvliet:
@jancborchardt 
"Btw before that's done in amy case - we should make sure we have all the svg sources or retain the high quality PNG still. Even if it's just dithering or so, future resolution improvements and such will bite us. ;)"
For that we have git... It keeps old images around ;-)

But yeah, makes sense for new images, to first commit full version and then optimize perhaps.

jancborchardt:
Yeah … using git to "have source files around" is just not discoverable. ;)

PR #583, opened 2017-09-07
Update page-homepage.php

jospoortvliet:
How about we link to this on the home page... The conf is old news ;-)

### DIFF
--- a/page-homepage.php
+++ b/page-homepage.php
@@ -12,9 +12,9 @@
 	<div class="container">
 		<div class="row">
 			<div class="col-md-6 toptext ">
-				<a class="announcement" href="https://nextcloud.com/blog/european-datacenter-is-no-solution/">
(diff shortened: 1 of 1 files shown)


jancborchardt:
We're not going to use this to announce every blog post now, right? ;)

Remember that the main claim, detail text, screenshot and call to action are the most important elements. Anything nonessential like this should be avoided.

It's fine for things like new releases and the conference, but that's it.

jospoortvliet:
Well, this is quite a major thing for customers etc, it certainly isn't "every blogpost" as we published more than a dozen since we did this one. If you have a better idea let me know, until then I think this is a good thing to promote. What else? Marius had some ideas... This is at least something our prospective customers should read.

jancborchardt:
Alright, your call. Agree the currently pinned tweet should be updated, maybe with the conf recap post?

jospoortvliet:
Allright, conference report is what it links to. Now we're already weeks after the conference, can somebody PLEASE just approve SOMETHING, ANYTHING that is better than our embarrassing "join the conference August 22-27" please? If you don't like this propose something better.

PR #591, opened 2017-09-20
Update page-events.php

jospoortvliet:
Add all-things-open

If approve, pls merge ;-)(


jospoortvliet:
There's no php in this so nothing that can truly break and it is a tiny change - going.

PR #605, opened 2017-09-28
Load fonts from website instead of loading them from Google

jancborchardt:
Not sure why we switched from serving these from our website itself to Google APIs. ;) It’s a bit undermining of our message. (Not sure if this already makes it all work, hope it’s a good step though.)


jospoortvliet:
We use CDN's for performance reasons... It makes our site slower for people who never visited it before. And slower loading = worse experience + worse ranking.

Just answering the 'why', I honestly find it a hard question what the smart thing to do is, but the new big refactor PR from paulo puts a lot more resources on CDN's and well, that's the same thing, right? So if we don't want that and accept that our site is slower, fine, but we have to decide.

I'm thinking the result of faster loading is more important than this anti-google thing tbh... it is used everywhere anyway.

jancborchardt:
> more important than this anti-google thing tbh... it is used everywhere anyway

I mean … that’s kinda the whole point. ;) Because it is used everywhere, people are being tracked everywhere. Just as if we had the Facebook-like button on every page – you know the problem.

jospoortvliet:
> I mean … that’s kinda the whole point. ;) Because it is used everywhere, people are being tracked everywhere. Just as if we had the Facebook-like button on every page – you know the problem.

Yeah, I get it. But isn't, say, cloudflare the same thing when it comes to tracking? Yet, if we disable it, loading our site from Australia will take, like, a minute if not more 😢 😭 
Perhaps the font is on another CDN, if we trust cloudflare more than Google (I could understand that). I'd be happy to have it replaced, then.

But removing Google would only make sense if we also get rid of ReCaptcha, I think, it's used on a dozen pages... We'd need an alternative that prevents shitloads of spam in a reliable way.

jancborchardt:
Ok so makes sense. One thing though is loading too many fonts - so I would do another PR with just the first commit, which removes Lato and the other thicknesses we don't use as far as I can see?

jospoortvliet:
That makes total sense, yeah.

Espina2:
We should use cdn as much as we can since, if you already download the files from any other website that use the same cdn it uses from cache. Another advantage is that cdns use http2 that can handle more requests at once, and we only use http1.

jospoortvliet:
Yeah, I think we established that ;-)

jancborchardt:
Closing in favor of this new PR just reducing the amount of fonts we load, further bettering performance: https://github.com/nextcloud/nextcloud.com/pull/611

PR #609, opened 2017-09-28
make E2e more prominent

jospoortvliet:
Do we want our end-to-end encryption to show up a lot more prominent? Link to it from our news teaser on the front page and from the menu...
What do you think? I'm not sure about it, but changing our home page link once a month to our 'biggest' announcement (we do them roughly every month anyway) is not so bad. And e2e is big enough for the menu, but the menu is already so big so ... pfff. 

Oh and 'why not wait until 13' -> that makes the menu no shorter (well, only temporary) and we decided to promote it so let's do it right ;-)

@nextcloud/website 

image

Thoughts?

### DIFF
--- a/templates/header-top-navbar.php
+++ b/templates/header-top-navbar.php
@@ -49,6 +49,7 @@ require get_template_directory().'/strings.php';
 					<ul class="nav__links ">
 						<li class="nav__item"><a href="<?php echo home_url('features') ?>"><?php echo $l->t('Overview');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('secure') ?>"><?php echo $l->t('Security');?></a></li>
+						<li class="nav__item"><a href="<?php echo home_url('endtoend') ?>"><?php echo $l->t('End-to-end encryption');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('sharing') ?>"><?php echo $l->t('Sharing');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('clients') ?>"><?php echo $l->t('Clients');?></a></li>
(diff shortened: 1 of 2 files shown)


templates/header-top-navbar.php
```diff
@@ -57,6 +57,7 @@
 						<li class="nav__item"><a href="<?php echo home_url('collabora') ?>"><?php echo $l->t('Online office');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('webrtc') ?>"><?php echo $l->t('Video chat');?></a></li>
 						<li class="nav__item"><a href="<?php echo home_url('outlook') ?>"><?php echo $l->t('Outlook integration');?></a></li>
+						<li class="nav__item"><a href="<?php echo home_url('sharing') ?>"><?php echo $l->t('End-to-end encryption');?></a></li>
```
jancborchardt: »sharing« seems to be the wrong link. ;)

Otherwise fine. Maybe even move it up to be directly behind »Security« since it kinda belongs to that?

jospoortvliet: Will fix that one. About moving it up, I did that initially but I just feel it looks odd because it is so long... a thought on shortening the name would be welcome, I couldn't think of anything.
image
what do you think, should I just get over it and move it up? Or keep it down...

jancborchardt: I’d say it’s fine, even so long. Otherwise we can call it »Encryption« and the header of the page can still be »End-to-end encryption«.

jospoortvliet: Yeah, if we call it encryption we have to add our server-side encryption there. I might do that later. Will put it up there for now, then you can approve this ;-)

jospoortvliet:
Jan, all good? Then pls merge!

PR #611, opened 2017-09-28
Load only the fonts we actually need, Open Sans Light & Semibold

jancborchardt:
Question is if we actually do need 300i, 400, 400i, 600i – @Espina2 @jospoortvliet?

Please review @nextcloud/website


jospoortvliet:
I can't judge if we use the other fonts but I doubt it, I never use a 'special' font anywhere. But I'd like to hear from @Espina2 - if he also doesn't know any instances we should merge this, it's a nice improvement!

Espina2:
@jancborchardt We use also regular. @jospoortvliet Did you not use italics on a page the have quotes from customers?

jospoortvliet:
@Espina2 ah you mean this page: https://nextcloud.com/migration
No, the quotes use bold.  But this page uses italics in the quote from NCC Group: https://nextcloud.com/secure/

jancborchardt:
Alright then this should work. I left in Regular as well as italic for all of Light, Regular and Semibold.

PR #631, opened 2017-10-16
fix & update order forms

jospoortvliet:
This should:
* fix the order form not working due to a reshuffling of javascript (sigh)
* fix the style of the contact, provider and order forms and their submission results.

Tested each form.


page-contactsubmit.php
```diff
@@ -19,13 +19,22 @@
 <?php
 if(isset($_POST['email'])) {
    function died($error) {
-    // error code goes here
-        echo "We are very sorry, but there were error(s) found with the form you submitted: ";
-        echo "<br />";
-        echo $error."<br />";
-        echo "Please go back and fix these errors.<br />";
-        die();
```
MorrisJobke: Half life time of this little issue here until it got found and used: 1 month and 4 days 🙈

jospoortvliet: what was the issue?!? Sorry, ... the 'die' broke the page design upon faulty submission.

PR #634, opened 2017-10-17
add cookie warning

jospoortvliet:
spectacle jl5819

@Espina2 will be proud of me, I did it all following the 'proper' js and css modules etc... ;-)

I know the js code isn't super and could be cleaned up a bit but it is good enough.

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -4590,6 +4590,9 @@ footer.page-footer {
     max-width: 600px;
     opacity: 0; }
 
+/*Colors*/
+/*Shadows*/
+/*Breakpoints*/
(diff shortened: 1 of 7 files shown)


jancborchardt:
Ugh, we have to do that now too? Damn. Just some points:

- the x button could use some padding, so it's a bit more centered and easily clickable
- the "that's fine" button should be closer to the text
- can you show a screenshot of how it looks on mobile? Should also only be one row

jospoortvliet:
* [x] gave X button padding
* [x] put button closer to text
* [x] shorten text and hide X button for mobile (X hides on screen width < 768px, the 'small break')

image

PR #648, opened 2017-10-31
Improve cookie warning

MariusBluem:


MariusBluem:
Please review! @jancborchardt @LukasReschke @NinaNC

jancborchardt:
Yeah, I guess it’s fine to get rid of the x. Or is it required @jospoortvliet?

MariusBluem:
Don't think so. If you do not agree ... you should close the window :)

PR #653, opened 2017-11-01
Refactor Code of Conduct

MariusBluem:
bildschirmfoto 2017-11-01 um 12 12 23


jospoortvliet:
- [ ] please let's not spend time translating this now. If a contributor can't read this, they can't really participate as we use English in all our communication as a community. The website is already badly translated, adding a HUGE amount of text right now seems a bad idea to me, slowing down efforts to reach 100% even more. Maybe, some day, so local communities can link to it - but not now, unless you do the translation to German of course, too 👅 
- [ ] GREAT image, really! But it is 1.54 MB, that shouldn't be more than 150-200KB ;-) can you adjust resolution, run it through https://tinypng.com/, etc until it is that big? Look at the other hea

MariusBluem:
> please let's not spend time translating this now. If a contributor can't read this, they can't really participate as we use English in all our communication as a community. The website is already badly translated, adding a HUGE amount of text right now seems a bad idea to me, slowing down efforts to reach 100% even more. Maybe, some day, so local communities can link to it - but not now, unless you do the translation to German of course, too 👅

Nobody wants to translate it right now 😝 But it should™️ be possible 😉

NinaNC:
Way better with a header indeed. 
But the dashes replacing the bullets look a bit weird.

MariusBluem:
> But the dashes replacing the bullets look a bit weird.

I agree. How to do dashes with our template in a proper way? Better than none :)

jospoortvliet:
You'll have to adjust the CSS to override the removal of disk bullets. I did it on another page but can't find it ATM...

jospoortvliet:
About the translation - by putting the translation in here you simply decrease the percentage of translated strings and if we decide to pay for translation (likely, as it doesn't get done) this page is likely to get translated (and that costs, well count, easily 30 cent per word). I really would prefer we just remove the translations for now from any page that shouldn't be translated, so we get our translation to 100% as soon as possible and at minimal cost and then we can finally deploy the German website. This simply increases the barrier to doing that, maybe not in theory but in practice: "can't we just ignore this page? Sure we can but we have to remember to do so".

Sorry, NACK on this from my side until translations are removed.

PR #656, opened 2017-11-08
iOS app is now free

NinaNC:
Changed the pricing of the apps


NinaNC:
was already done in another PR.

PR #664, opened 2017-11-14
Update README.cd

jospoortvliet:
Small typo at the Notes section about changing SCSS files. "Chaning" -> "Changing"

@jgorgulho not sure if you wanted to add other things? I created a PR because I thought this is worth adding and it's been 25 days since you made the change in this branch!

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,5 +31,5 @@ process easy and effective for everyone involved.
 ### Notes
 
 * Please don't commit straight into the master or live branches, these branches should remain as stable as possible, and changes should be discussed amongst the community.
-* If you are chaning SCSS files during development, following command for automatic compilation is recommended:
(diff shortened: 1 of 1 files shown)


jgorgulho:
Hi @jospoortvliet,
I only noticed the typo and got now it would be the only think I would propose to change.
Have I missed any step?

jospoortvliet:
Hi @jgorgulho you did the actually work by mading the change in a branch. Now we need to do the administrative part: others like me usually only see it when you create a 'pull request' like this one. Then it can be reviewed, discussed, merged and deployed!

So, I reviewed this, it looks good - let's merge it! 👍

jospoortvliet:
Thank you for your contribution and congrats for getting it in!

PR #692, opened 2017-12-10
Customer logos fix

MariusBluem:
as discussed @jospoortvliet

### DIFF
--- a/page-education.php
+++ b/page-education.php
@@ -45,7 +45,7 @@ require(["require.config"], function() {
 		<div class="col-sm-3 customer">
             <div class="customer-logo">
                 <div class="scaling-svg-container">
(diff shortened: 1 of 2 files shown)


page-education.php
```diff
@@ -45,7 +45,7 @@
 		<div class="col-sm-3 customer">
             <div class="customer-logo">
                 <div class="scaling-svg-container">
```
jospoortvliet: you should remove that, if this isn't a svg ;-) And the next div ofcourse

page-enterprise.php
```diff
@@ -170,7 +170,7 @@
 		<div class="col-sm-3 customer">
             <div class="customer-logo">
                 <div class="scaling-svg-container">
```
jospoortvliet: this too

PR #693, opened 2017-12-10
Add Tomaz and Dogan

jospoortvliet:
Adding Tomaz and Dogan. Also fixed where Jan & Joas live.


jospoortvliet:
@doganoo @tcanabrava pls review ;-)

PR #709, opened 2017-12-20
Linkfix

jospoortvliet:
Fix two links and add PicoCMS to the list of apps.


MariusBluem:
Killing the group folders is intended?

jospoortvliet:
yeah, it was intended - I simply removed the least interesting feature (which I had only added to make a round 6 in the first place) ;-)

jospoortvliet:
Ok with it now? I'm assuming this was all your issue with this so - I will merge & go...

PR #722, opened 2018-01-06
add compliance section to our home page

jospoortvliet:
Compliance is a key concern of customers so I added a section linking to our latest GDPR blog, our Healthcare page (HIPAA!), our auditing/monitoring page and our workflow/File Access Control page.

Here's a pic of what it looks like:
compliance


jospoortvliet:
I believe the Firewall icon could do with a slight adjustment (white behind the wall, perhaps?), thoughts?

The laptop icon (top-left) also could perhaps use some work... Not sure.

NinaNC:
I would replace "Compliance without interruption" by "Integrated compliance" or "compliant by design" (which is true since we integrate with existing systems. If they are compliant, we are). What do you think?

jospoortvliet:
The icon will be fixed in another PR I still have locally... 🙈

PR #734, opened 2018-01-11
Learn more ≠ Download

MariusBluem:
@jospoortvliet

### DIFF
--- a/page-homepage.php
+++ b/page-homepage.php
@@ -19,7 +19,7 @@
 				</a>
 				<h1 class="jumbotron--heading-1"><?php echo $l->t('Nextcloud - Protecting your data');?></h1>
 				<h2 class="jumbotron--lead"><?php echo $l->t('Building products that allow you to be productive without losing control');?></h2>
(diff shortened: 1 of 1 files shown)


page-homepage.php
```diff
@@ -19,7 +19,7 @@
 				</a>
 				<h1 class="jumbotron--heading-1"><?php echo $l->t('Nextcloud - Protecting your data');?></h1>
 				<h2 class="jumbotron--lead"><?php echo $l->t('Building products that allow you to be productive without losing control');?></h2>
-				<a class="button button--large button--arrow" href="<?php echo home_url('install') ?>" role="button" id="get-nextcloud-button"><?php echo $l->t('Learn more</a>');?>
+				<a class="button button--large button--arrow" href="<?php echo home_url('about') ?>" role="button" id="get-nextcloud-button"><?php echo $l->t('Learn more</a>');?>
```
splitt3r: The `</a>` shouldn´t be in the $l->t call.

jospoortvliet:
Sigh, I don't know what to do here. About isn't very useful. I just don't know what to link to...

I think I'm going to make an anchor and link to that, OK? So it will just scroll down...

PR #740, opened 2018-01-12
how about this?

jospoortvliet:
Houston, we have a problem: I am a sucky designer. I can't figure out how to make the list of features for the Android & iOS app look good... The initial look was this:
screenshot_20180112_103629

Then I tried this:

screenshot_20180112_103250

and now this:

screenshot_20180112_103155

Imho, it is all... meh.

Input, @nextcloud/designers perhaps?


jancborchardt:
How about basically taking the images as we have on the app store, and putting them as full-width sections underneath each other? Text is shown next to the phone.

jancborchardt:
Of course this will result in a very long page. If we don't want that, I would just literally take the images from the screenshots on the app store, and put 3 next to each other, maybe two rows. With the text above them (as actual written text, not image).

jospoortvliet:
Yeah, I could group it, put screenshots above, make it quite different than a single, stupid list. Probably the right thing to do, but more work 😭

jancborchardt:
@jospoortvliet yeah. Except that I could only recommend cutting down the text to like 3 items and making it larger. Or even just put 2-3 more screenshots down below? All step by step. :)

jospoortvliet:
Mwah, length of the page isn't a huge deal. Paulo suggested to go for the wide list we also used on https://nextcloud.com/talk/ and some other pages - that looks nice.

I'll do that, then add more screenshots. 🍰 eaten AND still here!

jospoortvliet:
Implemented what I described above. Feedback welcome! I think this is good enough, if you agree 👍 and let's get this live 🕺 
clients_ _nextcloud_-_2018-01-22_20 30 45

jospoortvliet:
sjees, rebase ruined that one... sorry.

jospoortvliet:
@jancborchardt what do you think? I know it isn't perfect but it's, I guess it is good enough...

jospoortvliet:
superceded by

PR #746, opened 2018-01-15
fix header for old pages

### LINKED ISSUE #736: Menu does not work on old pages like page-install
bildschirmfoto 2018-01-11 um 14 01 39

@jospoortvliet

jospoortvliet:
fixes #736 @MariusBluem pls test, I deployed on staging.

So that means - go to https://nextcloud.com/install?staging=true and see a better working menu.


jospoortvliet:
Works for me in staging in chromium and firefox... I'd rather get this fix live than wait ;-)

PR #780, opened 2018-02-05
Nextcloud 13 announcement

NinaNC:


NinaNC:
Yes, my bad, did two commits instead of one, I should have commented the old 12 stuff I was relying upon (I'm lazy and try not to type more HTML than needed ;) )

jospoortvliet:
It looks good - I'd like to merge.

Made a separate PR for the slideshow - normally I don't do that right away but I had some time in the train yesterday.

PR #785, opened 2018-02-06
Fix 13 release :)

MariusBluem:
This needs to be 13 - otherwise the docs URL will be generated wrong. The other one stays 13.0.0 for the download link :) Please review and deploy!!


strings.php
```diff
@@ -2,7 +2,7 @@
 // Strings used through nextcloud.com
 
 /* VERSIONS */
-$VERSIONS_SERVER_MAJOR_STABLE = '13.0.0';
+$VERSIONS_SERVER_MAJOR_STABLE = '13';
 $VERSIONS_SERVER_FULL_STABLE = '13.0.0';
 $VERSIONS_SERVER_MAJOR_DEVELOPMENT = '13.0.0';
```
MariusBluem: This should be 14 or empty now...

strings.php
```diff
@@ -2,7 +2,7 @@
...
 $VERSIONS_SERVER_MAJOR_DEV_DOCS = '13'; // Used in dev docs links
```
MariusBluem: Same here...

MariusBluem:
And please see my comments above... am I right?

strings.php
jospoortvliet: you are correct, this is wrong, our mistake.

MorrisJobke: Please keep it - we don't have 14 yet available on the docs server.

PR #790, opened 2018-02-11
first draft of media and advertising page

jospoortvliet:
bigmedia

Feedback welcome @mathiasconradt @NinaNC @MariusBluem 


tflidd:
Did you do something about the temp files Adobe creates? This was creating some issues in the past:
https://help.nextcloud.com/t/issues-syncing-large-photoshop-files/20530

It wouldn't be great if you sold it as "the" solution for media sharing and the first syncing already creates issues....

Also a question that pops into my mind: Delta-sync, or do files change so little that there is a real benefit of it?

MariusBluem:
I dont like the page name „bigmedia“. I think something like „agencies“ sounds netter ;)

mathiasconradt:
@tflidd Is it a real issue or only one user reporting this (the post on Stackoverflow is of 2014)? It's not on github, is it? I can only test with Photoshop Elements, and saving files there works ok. But they' usually not the largest. Should this not be handled/covered by chunked upload anyway?

@MariusBluem Agree. I didn't see "bigmedia" anywhere, but if it's the url then for SEO purposes, I would name it "media-and-advertisement-agencies", or if this is too long, "agencies" is ok, although it sounds a bit generic. Agencies can be everything, not just media & advertistment.

MariusBluem:
Pages like this, education, health etc. should be sub-pages IMHO:
nextcloud.com/health -> nextcloud.com/„casestudy“/health 
This would help keeping the page as organzied as possible with a clear hirachry:)

MariusBluem:
> @tflidd Is it a real issue or only one user reporting this (the post on Stackoverflow is of 2014)? It's not on github, is it? I can only test with Photoshop Elements, and saving files there works ok. But they're usually not the largest. Should this not be handled/covered by chunked upload anyway?

Anyway - something for the software code (Nextcloud/server) and not this homepage :)

mathiasconradt:
> nextcloud.com/health -> nextcloud.com/„casestudy“/health

I would make it "/industries/" then though, cause it's not really a case study.

tflidd:
> @tflidd Is it a real issue or only one user reporting this (the post on Stackoverflow is of 2014)? It's not on github, is it? I can only test with Photoshop Elements, and saving files there works ok. But they're usually not the largest. Should this not be handled/covered by chunked upload anyway?

I didn't follow up this issue so I don't know the current status. But as you mention these products explicitly, I thought you might have tested this (and people probably expect this). Chunked upload only breaks a large file into smaller pieces but does not check before, if the chunk at the receiving end is identical.

jospoortvliet:
I will rename it to 'agencies'. Bigmedia was indeed a work-around: couldn't use media, and 'advertising' for some strange reason gets blocked by my adblocker 🙈 

The subfolder thing I can do, too. Will put health and education there, too.

jospoortvliet:
WRT the issue that @tflidd linked to, https://help.nextcloud.com/t/issues-syncing-large-photoshop-files/20530 - we should certainly address this. @mathiasconradt can you test this and, if it is a real issue, bring this to the attention of engineering?

mathiasconradt:
I only have Photoshop Elements. Saving and syncing in general works fine with my previous tests for smaller files. Need to somehow create a file that's larger than 100MB first in order to test it properly, but sure, I can test it.

jospoortvliet:
Please do. It might indeed be that there are issues with files over 100mb (I doubt it matters from what application they come), and we should of course address that if that is the case.

mathiasconradt:
I think it's Adobe/Photoshop specific and the way temp files and renaming are handled by that, at least from the description in the linked help forum entry. Otherwise, there is already one open ticket about large files https://github.com/nextcloud-gmbh/customers/issues/277

jospoortvliet:
Thing is that most apps, certainly those that are cross-platform, deal with changing files that way: work in a temp file and rename it to the final file to 'save' it. So this will certainly hit more than Adobe users. Also, the issue you link to is about a 99GB file, that's an entirely different class (and type of error) I am sure... This problem seems more file locking related. Can you reproduce it?

PR #791, opened 2018-02-12
Legal page first draft

jospoortvliet:
A page for lawyers, notaries and law firms.

legal


tflidd:
The first thing I thought for this type of usage: Wouldn't it be great to be able to sign documents and/or files? (ref: https://help.nextcloud.com/t/feature-request-document-signing/13533)

tflidd:
Or some app to scan documents and manage them (automatically classed via OCR)?

It's also not clear, what Nextcloud support actually provides. Once you have Nextcloud support, do you define what kind of security is put in place and is it verified by Nextcloud (you probably won't do full security audits)? Imagine a company is hacked and the attackers got some data, it wouldn't be nice to learn that the encryption app wasn't enabled at all.

jospoortvliet:
@tflidd on the feature - yes, that is a nice request. Arthur worked on a 'I have seen it' check box thing, but of course full signing would be a step deeper. This is a typical feature for this market. We'd have to see if we can develop it, given enough demand from paying customers.

wrt the support we offer, we do offer a security review if a customer wants - though that costs extra. We do help them with advice around installation, making sure it is set up securely, though, and we offer the security scanner of course. We've been thinking about what else we could offer, ideas are welcome!

page-legal.php
```diff
@@ -0,0 +1,325 @@
...
+            <div class="ncc--brand--logo"></div>
+            <p class="ncc--brand--text">
+                <?php echo $l->t('NCC Group');?>
+            </p>
+        </div>
+    </div>
+    <div class="row feature-row">
+        <div class="col-md-10 col-md-offset-1 revealOnScroll feature--block">
+            <p class="section--paragraph revealOnScroll"><?php echo $l->t('Nextcloud offers first-in-class, third-party verified security backed by a USD 5000 Security Bug Bounty program protecting against unauthorized access.');?></p>
+            <a href="/secure" class="button button--blue button--arrow button--large"><?php echo $l->t('security in Nextcloud');?></a>
```
splitt3r: All the other buttons start uppercase first.

page-legal.php
```diff
@@ -0,0 +1,325 @@
...
+    </div>
+    <div class="row">
+        <div class="col-md-10 col-md-offset-1 revealOnScroll feature--block">
+            <p class="section--paragraph revealOnScroll"><?php echo $l->t('Legal documents are extremely sensitive. Clients care deeply about their privacy and keeping their data secure while regulators give heavy penalties for data leaks. Nextcloud offers you the tools to keep data compliant and safe.');?></p>
+            <p class="section--paragraph revealOnScroll"><ul>
+            <li><strong><?php echo $l->t('Legal compliance');?></strong><br/>
+            <?php echo $l->t('Federal data protection and GDPR-compliant, protecting data sovereignty, keep the server location in your country.');?></li>
+            <li><strong><?php echo $l->t('data security');?></strong><br />
+            <?php echo $l->t('Multiple levels of encryption (HTTPS/SSL/TLS, AES-256 or stronger, server-side and end-to-end encryption) ensure the highest level of data protection from hacking.');?></li>
+            <li><strong><?php echo $l->t('client- and case related authorization');?></strong><br/>
```
splitt3r: The first one is uppercase first the other ones aren´t.

page-legal.php
```diff
@@ -0,0 +1,325 @@
...
+            <i class="fa-tags fa"></i>
+        </div>
+        <div class="col-md-3 featureblock">
+            <p class=""><?php echo $l->t('Nextcloud offers an easy to use user interface which comes with powerful search functionality, trash and versioning, favorites, tags and more ways to quickly reach the files users need.');?></p>
+        </div>
+        <div class="col-md-1 featureblock">
+        <i class="fa-mobile fa"></i>
+        </div>
+        <div class="col-md-3 featureblock">
+        <p class=""><?php echo $l->t('Desktop and mobile clients for Windows/MacOS/Linux, Android and iOS complement the web interface, integrating in the file system. Nextcloud can also be reached through WebDAV.');?></p>
```
splitt3r: Only a personal preference but Windows/MacOS/Linux vs LDAP / Active Directory. I would also add spaces here.

mathiasconradt:
> Or some app to scan documents and manage them (automatically classed via OCR)?

Definitely useful in this field. Scanning software is already there on the market (Ephesoft, Kofax, any better scanner has WebDav interface on-board), can also use Scanbot, which can all be connected to Nextcloud out-of-the-box. Of coures, classification and tagging is another thing, requires a custom apps. File auto tagging wouldn't be sufficient, but this is all detail which can be discussed in a direct talk with prospects coming to the page.
General information that it's in general possible to connect to scanners, mobile scan apps, OCR, as it is mentioned under "Features" should be sufficient imho.

> It's also not clear, what Nextcloud support actually provides.

Not important on this kind of "landing page". Idea of this page is to get a first contact and in talks with potential users.

mathiasconradt:
> The first thing I thought for this type of usage: Wouldn't it be great to be able to sign documents and/or files? (ref: https://help.nextcloud.com/t/feature-request-document-signing/13533)

Agree, but I would leave this to products/app that already do this best. Makes imho no sense to develop what's already out, if we cannot make it at least as good as them, which would also need to include a mobile app, etc. I use DocuSign for example (and before SignEasy). 

I am just wondering why the mobile Nextcloud App (File App) doesn't register as a file source in the Android system, like Dropbox, Google Drive. etc. I will check if there is an open ticket about it, because that would definitely make sense.

In DocuSign, selecting a document to sign:

image

But also, this whole electronic signature topic for lawyers is kind of special anyway. BRAK already has  something for "real" electronic signatures in place https://www.bea-brak.de/xwiki/bin/view/BRAK/%2300115 but that would be way too complex too implement/connect to. We should leave that to law practice management solutions. 


On another note: the legal page currently doesn't say anything about connecting to existing systems, such as law practice management solutions. In Germany, that would be "ReNoStar" and "RA Micro" as the leading ones for SMEs.

jospoortvliet:
> On another note: the legal page currently doesn't say anything about connecting to existing systems, such as law practice management solutions. In Germany, that would be "ReNoStar" and "RA Micro" as the leading ones for SMEs.
They are mentioned in the section about OCR - just above the big blue  "try demo" and "contact sales" section.

page-legal.php
```diff
@@ -293,7 +293,7 @@
         <i class="fa-mobile fa"></i>
         </div>
         <div class="col-md-3 featureblock">
-        <p class=""><?php echo $l->t('Desktop and mobile clients for Windows/MacOS/Linux, Android and iOS complement the web interface, integrating in the file system. Nextcloud can also be reached through WebDAV.');?></p>
+        <p class=""><?php echo $l->t('Desktop and mobile clients for Windows / MacOS / Linux, Android and iOS complement the web interface, integrating in the file system. Nextcloud can also be reached through WebDAV.');?></p>
```
MariusBluem: MacOS -> macOS

PR #812, opened 2018-02-28
Added TLS-option for webchat-links to freenode

Somebodyisnobody:
I just changed the links to the freenode-webchat to https instead of http.
It's still implemented in https://github.com/nextcloud/nextcloud.com/blob/db1041d886a124d2e3a7c0a3b325134aea1b053e/page-box.php#L91

I ignored
https://github.com/nextcloud/nextcloud.com/blob/70440ace0f4bce9b3a680ade7b6fcd222f126bb1/page-install-backup.php#L62
because it's a backup-file.

### DIFF
--- a/content.xml
+++ b/content.xml
@@ -425,7 +425,7 @@
 		</wp:postmeta>
 		<wp:postmeta>
 			<wp:meta_key><![CDATA[_menu_item_url]]></wp:meta_key>
-			<wp:meta_value><![CDATA[http://webchat.freenode.net/?channels=owncloud]]></wp:meta_value>
+			<wp:meta_value><![CDATA[https://webchat.freenode.net/?channels=owncloud]]></wp:meta_value>
 		</wp:postmeta>
 	</item>
 	<item>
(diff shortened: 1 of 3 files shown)


content.xml
```diff
@@ -425,7 +425,7 @@
 		</wp:postmeta>
 		<wp:postmeta>
 			<wp:meta_key><![CDATA[_menu_item_url]]></wp:meta_key>
-			<wp:meta_value><![CDATA[http://webchat.freenode.net/?channels=owncloud]]></wp:meta_value>
+			<wp:meta_value><![CDATA[https://webchat.freenode.net/?channels=owncloud]]></wp:meta_value>
```
tflidd: Shouldn't it be the nextcloud channel instead of owncloud?
Where does this link show up?

MariusBluem: Nowhere ...

tflidd:
I can't find the promote page online. Is this page still in use?

MariusBluem:
No it is not. This is a legacy page (oc org / promote)

content.xml
Somebodyisnobody: Aaahm yeah... 😖  didn't recognize that because thought, that's (self-evident) already #nextcloud 🤣

tflidd:
> No it is not. This is a legacy page (oc org / promote)

So we keep it for some reason?

MariusBluem:
> So we keep it for some reason?

No. But we don't pay for storage on GitHub and maybe we need it in the future ... Will be removed after transfer to https://github.com/nextcloud/website is finished 😉

jospoortvliet:
Marius is right. 

Meanwhile, there's nothing really wrong with these changes, so let's just do it ;-)

jospoortvliet:
Thanks! I appreciate the attention to detail! Now people use a secure webchat, yay! :tada:

PR #817, opened 2018-03-06
Update page-events.php

jospoortvliet:
Adding event in SKÖVDE. Yes, that exists :rofl:


jospoortvliet:
Thank you for reviewing @tflidd really helps move things forward!

PR #828, opened 2018-03-15
Update page-changelog.php

jospoortvliet:
The pulls->pull should be fixed in the script, I'll do a PR @MorrisJobke


jospoortvliet:
Thanks for reporting & checking @chk1

PR #830, opened 2018-03-16
note (twice!) that Nextcloud is NOT a mail server

jospoortvliet:
screenshot_20180316_172153
screenshot_20180316_172119

### DIFF
--- a/page-groupware.php
+++ b/page-groupware.php
@@ -69,6 +69,8 @@
 			<li>Easy-to-use folder management</li>
 			<li>Supports end-to-end encryption using OpenPGP</li>
 			</ul>
+			<br>
+			<small><strong>Note:</strong> Nextcloud Mail is a client for POP or IMAP servers;<br>Nextcloud does not include a mail server.</small>
 			');?></p>
         </div>
 	</div>
@@ -94,7 +96,7 @@
 			<p class="section--paragraph"><?php echo $l->t('Keep your relations organized');?></p>
 			<p class="section--paragraph"><?php echo $l->t('<ul>
 			<li>Track birthdays of your contacts</li>
-			<li>Share your addressbooks with your team</li>
+			<li>Share your address books with your team</li>
 			<li>Sync your contacts with phones and other devices</li>
 			</ul>
 			');?></p>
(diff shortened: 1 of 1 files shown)


tobiasKaminsky:
I would also put this (again) in the bullet points, as (I guess) many people only read bullet points to get an impression and do not read the text above it.
(Maybe there the term "client" as suggested is a good idea to not repeat oneself)

jospoortvliet:
Where exactly would you put it as third place? What bullet list do you mean?

tobiasKaminsky:
Hehe..sorry I overread the "Note" part beneath the three points.
So my only idea is to use "Nextcloud Mail Client" instead of "Nextcloud Mail".

jospoortvliet:
Well, I considered that already, but we have:
Nextcloud Mail
Nextcloud Calendar
Nextcloud Contacts
(and later Nextcloud Notes, Nextcloud Deck etc)
Adding 'client' to one of them is pretty ugly... So I'd rather avoid that.

tobiasKaminsky:
Yeah, came also into my mind. Then let us hope that this is enough ;-)

page-groupware.php
```diff
@@ -94,15 +96,15 @@
...
+			<li>Share your address books with your team</li>
 			<li>Sync your contacts with phones and other devices</li>
 			</ul>
 			');?></p>
         </div>
 	</div>
 	<div class="row">
         <div class="col-sm-8 col-sm-offset-2">
-            <p class="section--paragraph revealOnScroll text-center"><?php echo $l->t('Nextcloud Groupware relies on native IMAP and SMTP capabilities. It fluently integrates into your existing infrastructure while also providing its own.');?></p>
+            <p class="section--paragraph revealOnScroll text-center"><?php echo $l->t('Nextcloud fluently integrates into your existing infrastructure while also providing its own. Note that Nextcloud Mail is a client for POP or IMAP servers; Nextcloud does not include a mail server. ');?></p>
```
MariusBluem: POP and/or IMAP-infrastructure but does not include its own mail server.

jospoortvliet: Done ;-)

PR #837, opened 2018-03-24
add an example of using Mautic to our privacy policy

jospoortvliet:
Transparency is good, right? I don't like that the default text, while, in all its fancy words, saying we do track users, is so vague. So adding a clear example & a mention of Mautic seems fair and nice.


chk1:
I have two small suggestions:

>For example, we remove users with an @gmail address and would keep SIEMENS mail addresses.

I would reword this slightly (btw. Siemens does not use all-caps in their name):

>For example, we remove users with an @gmail address and keep obviously business related addresses (such as @siemens.de or @redhat.com).

And the following (addition in bold):

>We then might follow up on a End-to-end Encryption whitepaper with an offer to download our Server-side Encryption whitepaper; or **a follow-up email suggesting to** read a new blog about encryption we wrote.

[Piwik changed their name to Matomo](https://matomo.org/blog/2018/01/piwik-is-now-matomo/) at the beginning of 2018, that name is mentioned a few times in the Privacy Policy.

tflidd:
:+1: to make such legal text more understandable.

However, by saving the e-mail addresses and supposing a private usage, you are also tracking these users. How about a check-box for private and business usage? This way you can still runs stats saying that whitepaper x is very interesting for private users as well. You keep private addresses only a very short time in order to track abuse or whatever.

And for business users, perhaps already put an asterix and show on the buttom of the page a small explanation for the use of the e-mail address. For some it might be quite interesting to leave a contact, e.g. they are evaluating different cloud solutions, e.g. they are interested in white papers about windows integration, they would be happy to know that there will soon be a Windows application to mount storage as network drive. Or you see it is an university interested in external storage, you already have similar customers and something is in the pipeline.

jospoortvliet:
@tflidd 
> However, by saving the e-mail addresses and supposing a private usage, you are also tracking these users.
I automatically delete them from our Mautic database, after sending the white paper of course. 

> How about a check-box for private and business usage? This way you can still runs stats saying that whitepaper x is very interesting for private users as well. You keep private addresses only a very short time in order to track abuse or whatever.
I don't want to complicate the form more (or rather, there's other stuff I'd ask first) and I don't want to track private users at all. And we'd have to rewrite Mautic to be able to track statistics on users without keeping them in the database, that's not really going to happen, sorry :smile: 

> And for business users, perhaps already put an asterix and show on the buttom of the page a small explanation for the use of the e-mail address. For some it might be quite interesting to leave a contact, e.g. they are evaluating different cloud solutions, e.g. they are interested in white papers about windows integration, they would be happy to know that there will soon be a Windows application to mount storage as network drive. Or you see it is an university interested in external storage, you already have similar customers and something is in the pipeline.
Well, if they download a white paper I guess they just signalled they are interested. If they then don't want further follow up they can unsubscribe and they'll be blacklisted - no mails, ever...

jospoortvliet:
I've implemented the feedback I got from @chk1 - thanks! Any other concerns or ideas for improvements? If not I'd like to get this live, we can always do another PR to improve it further...

MariusBluem:
@acme @example-ltd @placeholder-corp instead of @siemens and @redhat ... add @icloud to not just mention a single home use provider (@gmail) ;)

jospoortvliet:
@MariusBluem done

PR #838, opened 2018-03-27
improve appliance page

jospoortvliet:
1. Link to appliances, not forum, when talking about Windows options
2. Improve texts/separation for Univention & tech&me VM (to discuss further at hackweek)
3. Add docker & snap image. Both very well maintained for a long time, so trustworthy enough, plus we note they are community maintained.

The new/updated page:
screenshot_20180327_144347

Whaddayadunk?

### DIFF
--- a/install-instructions.php
+++ b/install-instructions.php
@@ -83,34 +83,34 @@
                                 <div class="col-md-6">
                                     <h5><?php echo $l->t('Home user appliance');?></h5>
                                     <p><?php echo $l->t('<a target="_blank" href="https://www.techandme.se/">Tech and Me</a> maintains a VM designed to be the easiest way for less technical users to get Nextcloud up and running. It builds on Ubuntu Linux and is fully set up and configured with a secure SSL/TLS connection.');?></p>
-                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.techandme.se/nextcloud-vm/">home user appliance</a>');?></p>
+                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.techandme.se/nextcloud-vm/">Get home user appliance</a>');?></p>
                                     <p><?php echo $l->t('Find <a class="hyperlink" href="https://github.com/nextcloud/vm">source here</a>.');?></p>
                                 </div>
                                 <div class="col-md-6">
                                     <h5><?php echo $l->t('Enterprise appliance');?></h5>
                                     <p><?php echo $l->t('Nextcloud GmbH maintains a free appliance built on the Univention Corporate Server (UCS). It includes LDAP and user management as well as optional Collabora Online integration.');?></p>
-                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.univention.com/products/univention-app-center/app-catalog/nextcloud/">enterprise appliance</a>');?></p>
+                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.univention.com/products/univention-app-center/app-catalog/nextcloud/">Get enterprise appliance</a>');?></p>
                                     <p><?php echo $l->t('Find <a class="hyperlink" href="https://github.com/nextcloud/univention-app">source here</a>.');?></p>
                                 </div>
                             </div>
                             <div class="row">
                                 <div class="col-md-6">
-                                    <h5><?php echo $l->t('Docker');?></h5>
+                                    <h5><?php echo $l->t('Docker image');?></h5>
                                     <p><?php echo $l->t('Several Nextcloud community members maintain a Docker image. It supports a wide range of architectures and releases, features Apache/FPM, NGINX, various databases and more.');?></p>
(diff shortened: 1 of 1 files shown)


ngulden:
@jospoortvliet : My thoughts on this:

* shorten the headline for the enterprise appliance to just "Enterprise appliance". This should keep the headline on one row and it matches the button content.
* Change the appliance headlines to "Home User Appliance" and "Enterprise Appliance"
* Label the appliance buttons the same as the headline

The separation makes it now more explicit and the button labels tell the user much better, what he will get. Let's try that and see how it performs.

MariusBluem:
- Buttons should be labeled with „Get it“

jospoortvliet:
@MariusBluem the issue we tried to address was that people tend to just read the buttons and click. Making them more descriptive solves that ;-)

I think it fits with the general Nextcloud UI where we do the same - not OK and Cancel but 'Move' for example...

MariusBluem:
So then lets go with "**Get** enterprise appliance" and "**Get** snap" ... instead of just "Enterprise appliance" just that its crystal clear what the button actually does 🤗

install-instructions.php
```diff
@@ -78,22 +78,42 @@
...
-							<p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.techandme.se/nextcloud-vm/">Tech and Me VM images</a>');?></p>
-							<p><?php echo $l->t('Find <a class="hyperlink" href="https://github.com/nextcloud/vm">source here</a>.');?></p>
-							<br>
-							<h5><?php echo $l->t('Small and medium enterprise appliance');?></h5>
-							<p><?php echo $l->t('Univention offers an appliance with Nextcloud (maintained by Nextcloud) built on the Univention Corporate Server (UCS).');?></p>
-							<p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.univention.com/products/univention-app-center/app-catalog/nextcloud/">UCS Appliance download</a>');?></p>
+						<div class="col-md-8">
+                            <div class="row">
+                                <div class="col-md-6">
+                                    <h5><?php echo $l->t('Home user appliance');?></h5>
```
MariusBluem: appliance -> VM

install-instructions.php
```diff
@@ -78,22 +78,42 @@
...
+                                <div class="col-md-6">
+                                    <h5><?php echo $l->t('Enterprise appliance');?></h5>
+                                    <p><?php echo $l->t('Nextcloud GmbH maintains a free appliance built on the Univention Corporate Server (UCS). It includes LDAP and user management as well as optional Collabora Online integration.');?></p>
+                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.univention.com/products/univention-app-center/app-catalog/nextcloud/">enterprise appliance</a>');?></p>
+                                    <p><?php echo $l->t('Find <a class="hyperlink" href="https://github.com/nextcloud/univention-app">source here</a>.');?></p>
+                                </div>
+                            </div>
+                            <div class="row">
+                                <div class="col-md-6">
+                                    <h5><?php echo $l->t('Docker');?></h5>
```
MariusBluem: Docker image

install-instructions.php
```diff
@@ -78,22 +78,42 @@
...
+                                    <p><?php echo $l->t('Several Nextcloud community members maintain a Docker image. It supports a wide range of architectures and releases, features Apache/FPM, NGINX, various databases and more.');?></p>
+                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://hub.docker.com/_/nextcloud/">Docker image</a>');?></p>
+                                    <p><?php echo $l->t('Find <a class="hyperlink" href="https://github.com/nextcloud/docker">source here</a>.');?></p>
+                                </div>
+                                <div class="col-md-6">
+                                    <h5><?php echo $l->t('Snap');?></h5>
```
MariusBluem: Snap package

install-instructions.php
```diff
@@ -78,22 +78,42 @@
...
+                                    <p><?php echo $l->t('Canonical and the Nextcloud community maintain a Nextcloud Snap, including release channels and quick and easy deployment for easy home use.');?></p>
```
MariusBluem: Link to Canonical (you link to Tech and Me too 😉)

install-instructions.php
jospoortvliet: I just changed that to Appliance to make things more consistent ;-)

Home User VM
vs
Enterprise Appliance

is weird. Either they are both VM's or both Appliances. Technically, both are both (...) so let's keep the term Appliance, sounds more serious :D

jospoortvliet:
I agree with Marius, we can't explain docker vs snap etc here. People should know what they are looking for - or look in our docs (yeah, there we can improve the explanations, too).

jospoortvliet:
@MariusBluem pls have a look & approve if it is OK, so we can test this until the hackweek at least and see how it works...

install-instructions.php
MariusBluem: If home user „appliances“ should sound „serious“ is another topic ...

jospoortvliet: true, true, I just want to be consistent here...

PR #842, opened 2018-04-04
Add section on DICOM viewer

jospoortvliet:
Would also love to hear from @Jurrie @coliff if they see any typos or other issues!

screenshot_20180404_211943


splitt3r:
That doc has a lot of plugins installed on his work pc :grinning:

Jurrie:
I don't see any typos :) I do see a lot of changes merely because of formatting. What's Nextcloud's take on this? (Personally, I favour strict formatting rules, as they keep pull requests from cluttering.)

jospoortvliet:
@splitt3r
* yeah... that image isn't entirely 'real', you know how it is with stock pics :paintbrush: 
* will change the app store capitalization.
* I have some trouble finding the right term. 'community' sounds just a tad too much like a home user/student or 'just volunteer' and @ayselafsar is a professional developer with loads of experience in this area... This viewer isn't a toy. Now technically, that's still 100% community, no doubt about it, but I still feel a little uncomfortable with the term here, which is why I talked around it, simply not claiming anything about who did it. Suggestions on better words?!? "third-party developed" maybe? Or am I just being difficult here?

@Jurrie two things:
* in the .css file, lots is changed because it's generated from the scss
* I'm an elephant in a porcelijn store to use a Dutch saying :yum: I am just quite clueless with this stuff and not disciplined at all... Sorry.

PR #844, opened 2018-04-06
Update job text to differentiate sales from account mgmt

jospoortvliet:
Based on a discussion with Andreas - Account Management is, in the software business, a sales job, but it is a bit confusing. So we split into two: account management and sales rep. We do need both and people who are interested in either (or both) should apply...


jospoortvliet:
Thanks!

PR #848, opened 2018-04-11
add SME

enoch85:
What do you think @jospoortvliet?

### DIFF
--- a/install-instructions.php
+++ b/install-instructions.php
@@ -81,9 +81,9 @@
 						<div class="col-md-8">
                             <div class="row">
                                 <div class="col-md-6">
-                                    <h5><?php echo $l->t('SME and home user appliance');?></h5>
+                                    <h5><?php echo $l->t('Home/SME appliance');?></h5>
                                     <p><?php echo $l->t('<a target="_blank" href="https://www.techandme.se/">Tech and Me</a> maintains a VM designed to be the easiest way for less technical users to get Nextcloud up and running. It builds on Ubuntu Linux and is fully set up and configured with a secure SSL/TLS connection.');?></p>
-                                    <p><?php echo $l->t('<a class="btn btn-lg btn-primary" href="https://www.techandme.se/nextcloud-vm/">Get SME and home user appliance</a>');?></p>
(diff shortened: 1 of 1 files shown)


tflidd:
Or use something like single server appliance, so people can decide on their own if that is enough for their use-case. If you use small NC setups for each department, you could use this image on several vservers in large companies as well (and use federated sharing).

enoch85:
@tflidd This is not clustered per say, but you could cluster it if you have the know-how. As both you and me know a VM is like a server and you can add as many CPUs, RAM, and HDD as you want. I asked @jospoortvliet beacuse we discussed over email how we should present the Tech and Me appliance.

1. I understand that Nextcloud wants to make money of Enterprise customers
2. At the same time my VM could be used in enterprises as well. 
3. Though I'm not interested in Enterprise customers right now, but at the same time I don't want to "talk down" my own product that I've been working on since 2014, which by now also is a brand.

tflidd:
I meant more having a load balancer, different instances connecting to a database-cluster etc. There your image is probably not very helpful and it's probably not your target audience. So you want to run everything inside this VM (with optional external storage and external authentication). Not that it is impossible to do more and extend it. For me the main purpose is for people who want to run NC in a virtual machine, this is more or less a plug&play solution.

enoch85:
>  having a load balancer, different instances connecting to a database-cluster etc.

@tflidd So you mean that the Univention appliance does all that? In that case I rest my case.

tflidd:
> @tflidd So you mean that the Univention appliance does all that? In that case I rest my case.

I know vmware images because I've already worked with it. No idea how Univention works and the advantages compared to other solutions (don't even start with docker and snap). The overview page does not provide enough information for me to make a choice, I would have to look into the details about the difference between these solutions in principle and then specific differences in Nextcloud (update procedures, migration to other platforms...). Perhaps I should know all this.

jospoortvliet:
I agree we should try to change the wording, thing is, this gets really really long... Maybe "home/SME appliance"?

enoch85:
Fine with me!

PR #849, opened 2018-04-12
update about page

jospoortvliet:
OK'ed by Frank & @mathiasconradt


splitt3r:
The links https://nextcloud.com/?p=1222 and https://nextcloud.com/?p=1343 don´t work

PR #852, opened 2018-04-20
add language choice buttons

jospoortvliet:
screenshot_20180420_225916
screenshot_20180420_225856
screenshot_20180420_225945

Can go live as soon as the last 50 strings are reviewed on Transifex & deployed. TODO:
* [ ] about page
* [ ] files page
* [ ] install page
* [ ] Nc Box page

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -189,12 +189,12 @@ th {
 
 html {
   font-size: 10px;
-  -webkit-tap-highlight-color: rgba(0, 0, 0, 0); }
+  -webkit-tap-highlight-color: transparent; }
 
 body {
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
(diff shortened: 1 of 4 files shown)


MariusBluem:
Could we only show the non-selected language as button? (If English - show German, if German - show English)

MariusBluem:
Do we have a flag which covers american and british english with only one button?

Example: https://commons.m.wikimedia.org/wiki/File:English_language.svg

jospoortvliet:
1. how is your PHP? Cuz no, I can't do that.
2. I could put a different flag up, sure, but I think this is good enough... Everyone will understand that these countries speak roughly the same language ;-)

MariusBluem:
1. PHP - what is that???!!! ;) ;)
2. Fine.

Otherwise: 👍

PR #855, opened 2018-04-24
Nextcloud at home - wording

MariusBluem:
@jospoortvliet @shironextcloud


jospoortvliet:
screenshot_20180424_180104

I am a little torn - it is quite long, not per-se TOO long but in German, it probably will:
For families, students & individuals
Für Familien, Studenten & Einzelpersonen
(google translate)

But besides that, yeah, it does make sense. It is more friendly. Come up with a shorter text for German and we merge ;-)

MariusBluem:
"For families, students & you" - "Für Familien, Studenten & dich"

PR #859, opened 2018-05-04
update to Nc 13 state.

### LINKED ISSUE #847: Adjust threat model
Your threat model at https://nextcloud.com/security/threat-model/ is sometimes a bit outdated as it seems. E.g. AFAIK in NC 11 or 12 you've integrated brute force protection, so 
> we currently expect people to protect their instance using measures such as fail2ban. We do have a native anti-bruteforce protection.

is no longer true.

Same for the Android content provider. I am not sure whether this is the "Open file" -> "Nextcloud" thing, but if it is that, this is also no longer true:
> (since no Content Provider is yet implemented).

Of course, you should describe there that downloaded files are saved on your sd card (or other storage) and are accessible by all apps.

Also as for the NC 13 you might add something about the e2e crypto there. (or link to your whitepaper or whatever)

jospoortvliet:
fixes #847

Need input on the Android app section. @mario @tobiasKaminsky @MorrisJobke @rullzer 

### DIFF
--- a/page-threat-model.php
+++ b/page-threat-model.php
@@ -15,8 +15,11 @@
 <p>We do consider local mounted storage systems as trusted, so if a symlink or something else is configured on the external storage the Nextcloud server will follow it with the web server privileges.</p>
 <p>For this reason we do recommend administrators to only use the external storage mount for ultimately trusted content.</p>
 
-<h3>Encryption</h3>
+<h3>Server-side encryption</h3>
 <p>Nextcloud can be configured to encrypt data at rest. In this scenario we do prevent against storage administrators mainly, we are aware that a Nextcloud administrator could still intercept the user password to manually decrypt the encryption key. We do thus only consider attack scenarios bounty-worthy if they include external parties.</p>
+
+<h3>Client-side encryption</h3>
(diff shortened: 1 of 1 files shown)


page-threat-model.php
```diff
@@ -29,7 +32,7 @@
 <p>At the moment we consider version disclosure an accepted risk as an attacker can enumerate service versions using other means as well. (e.g. comparing behaviour)</p>
 
 <h3>Attacks involving other Android apps on the device</h3>
-<p>We do consider attacks involving other Android apps on the device as minimal risk, also especially considering that the Nextcloud Android apps stores synced files locally accessible on the device. (since no Content Provider is yet implemented).</p>
+<p>We do consider attacks involving other Android apps on the device as low or medium risk. Downloaded files are saved on your sd card (or other storage) and are accessible by all apps.</p>
```
mario: Stored files can be hidden from other apps if appropriate storage option is selected inside the app.

If you have a compromised phone and a compromised pp, you're probably screwed though. We cannot protect against this.

page-threat-model.php
```diff
@@ -15,9 +15,12 @@
 <p>We do consider local mounted storage systems as trusted, so if a symlink or something else is configured on the external storage the Nextcloud server will follow it with the web server privileges.</p>
 <p>For this reason we do recommend administrators to only use the external storage mount for ultimately trusted content.</p>
 
-<h3>Encryption</h3>
+<h3>Server-side encryption</h3>
 <p>Nextcloud can be configured to encrypt data at rest. In this scenario we do prevent against storage administrators mainly, we are aware that a Nextcloud administrator could still intercept the user password to manually decrypt the encryption key. We do thus only consider attack scenarios bounty-worthy if they include external parties.</p>
```
tflidd: In the docs we say that the server-side encryption was designed for data on external storages. And bounty-worth attack scenarios must not involve an admin/root account of the Nextcloud server (either decrypt files directly without keys or find a way through other non-admin user accounts).

jospoortvliet:
we OK now? Pls approve :D

tflidd:
Just about the brute-force protection, a way around would be to use the TOR browser. Perhaps we need to be a bit more specific what we would consider a way around. If I manage the bruteforce protection to block access from any ip, I won't gain access but somehow I broke the feature...

PR #862, opened 2018-05-10
Update page-box.php

jospoortvliet:
Adding this as an alternative. Not super duper prominent yet, we can do more once the fundraiser is done and the device is on the general market. @miska FYI

Good stuff, btw :rocket: 

Small change but I'm very happy with this project!


MariusBluem:
Simply add it to page-devices ... unrelated to Nextcloud Box 🤨 You could say, that you can find other devices (including Syncloud, Spreedbox, ...) on the devices-page ...

miska:
Looks good to me

jospoortvliet:
@MariusBluem I thought about that, yes. But most people come FROM that page. I will do a PR later to add the MOX to that page... I want to do this as a temporary fix. OK with that?

tflidd:
Really interesting project! Good catch, hopefully they succeed...

jospoortvliet:
Ok, did what @MariusBluem suggested ;-)

This is what it looks like:
screenshot_20180510_120808
screenshot_20180510_120836

PR #863, opened 2018-05-10
Load Google Fonts faster with Preconnect

coliff:
Load the Google font files faster by adding the preconnect hint
The preconnect hint is supported by Chrome, Opera, Firefox and Android browsers

REF: https://www.cdnplanet.com/blog/faster-google-webfonts-preconnect/


jospoortvliet:
Good with me :+1: 

though we should stop loading from CDN's and just put everything on our own site at some point... Perhaps in our new website!

Thanks, @coliff

PR #867, opened 2018-05-20
Removed planet-part / typo phpList

MariusBluem:


MariusBluem:
> We only have redacted blogs from Nextcloud, no aggregation with other blogs any more?

For now, Yes. Our new blog is not working with the old „Nextcloud Planet“ blog aggregation. This may change with the new website (https://github.com/nextcloud/website) ;)

jospoortvliet:
Mja, it is more correct right now, indeed - but this is essentially a bug. And I prefer to already tell ppl about this too much than risk forgetting to add a disclaimer when we re-add the blogs...

Can we please keep that sentence? Then it's good!

jospoortvliet:
> I'm ok with keeping it if it is scheduled in a foreseeable  future (a couple of months). If you think it's rather a year or more, it  could cause confusion because it mentions features that are not  available at all.

I'm hoping we can continue with the new site, soon - yes. Should be months, certainly not be months!

PR #868, opened 2018-05-23
GDPR info

jospoortvliet:
* GDPR and RH partnership press releases
* FAQ item for Finance page & GDPR link in header


jospoortvliet:
Both points agreed. It IS translatable - please translate on transifex!!!

PR #872, opened 2018-05-30
Update IRC links to Riot and simply call it "chat" for better usability

jancborchardt:
Please review @jospoortvliet @danxuliu :)


jancborchardt:
Thanks! :) Yep - at this point we need to grow beyond people knowing IRC. People familiar with it but using another client can still figure it out easily.

PR #874, opened 2018-05-31
show yt imgs behind GDPR warning

jospoortvliet:
crazy shit, but it works *mostly*: if a YT image with the same YT hash as the video is in the /media folder, it shows it as background. Makes the videos look a lot less nasty but doesn't work when the page is in a subfolder (like our verticals) unless the yt images are, too... which I'll do on the server because I want to show them, obviously. Nasty but it works.

Oh and: done by editing minified JS and CSS. I know. Terrible, but it works.


jospoortvliet:
before:
screenshot_20180531_215904
after:
screenshot_20180531_215745

It is done in a way that it works on the blog too... That is the major 'feature', actually, make our blog GDPR compliant.

jospoortvliet:
I have uploaded thumbnails of the YouTube videos to our website, once this gets merged and deployed it should work.

jospoortvliet:
Tested it on a different system with different wordpress setup & after grabbing background images.

Checked all our pages with a YouTube video, it is working surprisingly well ;-)

PR #876, opened 2018-06-03
Gdpr tuning

jospoortvliet:
Hi awesomes!

I created a new GDPR page, besides a refactor the compliance section into a separate php file.

Love to hear what you think of the page!

gdpr-tinypng


jasonbayton:
The title is not clear due to the busy background image, also "data" is not capitalised about half way down

jospoortvliet:
Fixed the capitalization. Better image:
screenshot_20180604_103414

jasonbayton:
Awesome. It does go on quite a bit and likely duplicates the message from other pages, but if that's the intention then no worries. Last niggle - I think "self hosted" should be "self-hosted", what do you think?

jospoortvliet:
Agreed :D

Issue #864, opened 2018-05-10
Nextcloud Box page clarification


tflidd:
On the Nextcloud-Box page I found some things to clear up (https://nextcloud.com/box/):

> The software for the box as well as similar solutions are still being developed by Nextcloud contributors. You can use the Ubuntu Snap image for easy installation on small and large devices that already have an operating system. Alternatively you could use the NextcloudPi image, a full operating system aimed at the Raspberry Pi and other boards.

- it sounds like you could use Ubuntu snap on any system. Is this true? Or do you need Ubuntu?
- Nextcloud-pi, I'd rather write "a fully pre-configured operating system", it is not an operating system itself.

MariusBluem:
- Snap is a standard (https://snapcraft.io) you can use on other Linux-based OSes too (SUSE, Debian, RedHat). We test the snap against Ubuntu, but I see no reason why it should not run on Debian for example :)

- Mind open a PR for this wording change (Nextcloud Pi)?

PR #878, opened 2018-06-05
Add new NextcloudPi URL

tflidd:
Small changes to wording, putting the new name and link of the project.

Fixes https://github.com/nextcloud/nextcloud.com/issues/864


jospoortvliet:
This name change is going to change again - given that 'NextcloudPlus' sounds more like a Nextcloud fork than a disk image for development boards, I had to ask to change it. Remember the CalendarPlus and ContactsPlus apps 'back in the day', forks of the originals? :cry:

jospoortvliet:
Oh there are more changes than the name-to-plus, sorry for missing that. I haven't seen them pick a final name, might stick with NextcloudPi - for now I guess we should keep that here...

MariusBluem:
Nextcloud Plus is not a good name, since as @jospoortvliet mentioned it sounds like a fork. I don’t want to think about discussions like „I am running Nextcloud - well, I am even running Nextcloud Plus“ ... it sounds like the Plus makes Nextcloud better by design, which is not the case. It’s just another way of running the same code base with an very undescriptive and misleading  trademark.

tflidd:
They decided to stay with the RaspberryPi name (https://help.nextcloud.com/t/feedback-welcome-naming-for-nextcloudpi-project/29356/8). They have a dedicated URL, I also changed the wording a bit.

jospoortvliet:
Perfect!

PR #880, opened 2018-06-07
update press mentions on about page

jospoortvliet:
Adding a varied selection of the articles by the press about us over the last 2 years.

Also remove the owncloud history part - nobody cares anymore, Nextcloud is about the future :rocket: 
about_ _nextcloud_-_2018-06-07_12 06 38


tflidd:
With "user focus" you focus on business users. However, a very strong point is the community. Nothing about that and it makes this product very different to competitors. For potential customers, it might be less important, that everybody can have a fully functioning product. But they might have advantages since many ideas and applications come from the community.

jospoortvliet:
Good point, yes, I will work some community in there.

jospoortvliet:
Made that change. If no other feedback, can I get a :+1: ? :smile_cat:

Issue #493, opened 2017-06-13
Description of each job offer is jumping when collapsing


ViRPo:
I noticed, that when you expand or collapse job position at /jobs, for a short while there is bigger gap between job name and job description while it's expanding or collapsing

It can be seen here: http://recordit.co/ibltd5nPBw
Or you can test it yourself: https://nextcloud.com/jobs/

I failed to find the cause of this, so for now I am submitting it as an issue, however it seems to be connected with class "collapsing"

jospoortvliet:
Yeah, this probably has to do with conflicting javascript from bootstrap and the custom scripts we have :(

@Espina2 am I right?

Espina2:
@jospoortvliet No this is just css.
The problem is on having margins on the paragraphs if you remove it will work just fine.

jospoortvliet:
Ah, that simple huh. @ViRPo wanna try to fix it?

Bentong021:
https://github.com/nextcloud/nextcloud.com/issues/493

FWirtz:
@Espina2 I removed the margins using dev tools - doesn't seem to fix the jumping for me though 😞

ViRPo:
@FWirtz this fixed it for me:
![image](https://user-images.githubusercontent.com/9433048/31225520-c9f717bc-a9d2-11e7-80bc-28c7596838c7.png)
Want to try to fix it inside code?

FWirtz:
... works for me now.
I am pretty sure I did just that though.
Sorry for the confusion.

Yeah, will do! Hope I can make it today, otherwise next week as I'll be away.

axlwaii:
I fixed that [here](https://github.com/nextcloud/nextcloud.com/pull/884) :)

PR #884, opened 2018-06-20
fix: fix jumping description of each job offer when collapsing,

axlwaii:
I came around [this](https://github.com/nextcloud/nextcloud.com/issues/493) and thought I fix it quickly :)

**What did I do ?**
I added the following styles to `legacy-fixes.scss` since the bug was located inside the `collapse`-class:
```scss
.collapse.in p:first-child,
.collapsing p:first-child {
  margin: 0;
}
``` 

**The Problem**
The class `collapse` is replaced with `collapsing` this in combination with the margin on the p-tag caused the bug.

fixes https://github.com/nextcloud/nextcloud.com#493


jospoortvliet:
You're my hero, I tried to fix this many times and failed. THANK YOU @axlwaii !!!

PR #887, opened 2018-06-21
Register bump

skjnldsv:
Fixes:
 - No js detection warning removal on internet explorer
 - Register page ie11 compatibility
 - OCSAPI header fix


skjnldsv:
> I can only assume it is good ;-)

You're too kind my good sir! ;)

PR #890, opened 2018-06-22
add more details and clarifications to privacy policy

jospoortvliet:
Cleaning up git branches, discovered I never pushed this... Sigh.


tflidd:
Updater app: You say what information is transmitted but is Nextcloud saving these data? And do you save the IP address, or asked differently, can you identify individual setups (you could perhaps via ip address, the cloud-id, which probably is unique, is not transmitted.

The part about the usage survey app is very unspecific. It is completely unclear if you can identify specific setups and also follow configuration changes. If you loose the data, is there a URL of all setups? Could someone obtaining this data, identify potentially outdated systems? Or identify specific users? If you provide more information about this app, you should link it here.
This app is very special because you want information and there is no or very little benefit for a user. With such a declaration, I won't consider using such an app.

You also say, that apps developed by Nextcloud do not send any data. There are very few apps published for Nextcloud GmbH but many more are available and used. I think, you should say that you can't take responsibility for 3rd party apps.

Is there a policy for apps in the app store, and if yes is it verified? I don't want an app which shows me date and time and secretly transmits my address book. It would be nice to have a policy that apps must transmit as little as possible personal data to provide their function. And also they must notify when they transmit something and what (at least when enabling it). In general when something is sent, it is logged somewhere and I can take a look at it on my personal page.

jospoortvliet:
Thanks. Yes, I have to handle the points you make.

PR #900, opened 2018-06-26
Fix broken link in pricing page

kevgathuku:
Fix link in pricing page to point to correct post.

- [ ] https://nextcloud.com/pricing/?staging=true

Fixes this link:
output

cc @jospoortvliet


jospoortvliet:
BTW our staging has been broken since we had to re-setup wordpress some time ago, and we plan to migrate away from WP so we won't fix it until we have.

But I test locally and so do some others. Appreciate the contribution!

jospoortvliet:
tested, deployed. Needed an additional fix and prompted me to also update the GDPR link <- very good!

kevgathuku:
Happy to help and nice to hear that it led to other fixes 😄

PR #906, opened 2018-07-06
Outlook add-in: add new features, update link

jospoortvliet:
screenshot_20180706_160437


MariusBluem:
Support for high DPI resolutions is not really a feature - it’s something which can be expected from every app in 2018 ;-)

tflidd:
> lets you pick local files and upload them to Nextcloud. It can provide a link where the recipient of the email can upload files to and enables customization (via html) of the inserted information.

Does the recipient not download the files? Or do you want to say that you can send a upload link as well?
And the customization is done by the admin, by the user who is sharing his files or the recipient???

You should decide which spelling of e-mail/email you prefer, both seem to be possible (https://en.wikipedia.org/wiki/Email) but I wouldn't change it within one article.

jospoortvliet:
Removed the hi-dpi thing. Rewrote the part about uploading files. Everything is email now, no e-mail.

PR #907, opened 2018-07-10
Translation design fix

jospoortvliet:
refactor the css and made it work for german translation - the pricing page there looks awful:
https://nextcloud.com/de/pricing

This fixes the height of each item to 50 px, making the line distance shorter so lines that have to wrap don't impact the size of the table making them uneven.

It also changes the button design to the new style, which actually accommodates longer texts better.

I snuck in a tiny fix for a missing YouTube video on the Files page.


jospoortvliet:
Before:

screenshot_20180710_112220


After:
screenshot_20180710_112519

PR #912, opened 2018-07-16
Add two quotes to migration page

jospoortvliet:
Thought it would be nice to put some quotes in there...
screenshot_20180716_220748


tflidd:
Cloudwards says: "many users also report Nextcloud is much more stable than ownCloud"
It's a bit like the famous "people say that ..." where you can fill in anything you want. "Less feature rich and less developed", they could turn it around and say that they prefer to have stable releases rather than putting a ton of features in it that are less tested. In the end, everybody has to check the feature list and decide what's important. And instead of loosing too much time in ranting at ownCloud, it's better to convince customers to move away from proprietary solutions.

You had this article about TU Berlin where they reduced the database load impressively from one version to the other, that is a real experience and something you can proof. That makes people want to use Nextcloud, try out and test a new version.

jospoortvliet:
@tflidd you are right on both accounts, of course.

Then again, twitter is full of nice messages about us, I would love to highlight that. But maybe this isn't the best way.

And yeah, oC really likes justify their lack of development by claiming they are focusing on stability... While I guess most people realize it is an excuse, you're right it fits that story. Then again, they say that Nc is more developed - that includes stability I would say.

But maybe I should just put a list of "what others say" on the bottom with links to these guys, the TU etcetera. Migration is still an issue and people still ask about it so I do want to keep this page maintained.

jospoortvliet:
Ok, added links to more whitepapers etc - so putting it live... Small steps. At the moment, with the shrinking user base of ownCloud, migration isn't as big an issue so I don't want to put in more effort, at least until we've made the migration easier. It is a pita these days... We have to fix that first.

PR #922, opened 2018-08-06
Fix apps display on low width devices

skjnldsv:
Before
capture d ecran_2018-08-06_14-15-08

After
capture d ecran_2018-08-06_14-15-02


tobiasKaminsky:
:+1: 
I can test it once it is deployed, but looks good so far from the screenshots.

jospoortvliet:
deployed.

PR #944, opened 2018-08-30
Reordered devices by 1) available, 2) announced, 3) not available, and added recently announced UBOSbox devices

jernst:
I was unable to set up a staging nextcloud.com site per the README.me instructions (they seem to be incomplete and are a bit "non-standard" :-)), so this may have broken links etc.


jospoortvliet:
Looked good, put live.

PR #947, opened 2018-09-03
Make /devices page more readable and understandable; redirect from /box

jernst:
I hope I got the "tags" reasonably correct; I think it would be of help to the user to understand whether a device requires assembly, for example.

Also, /box does not seem to be referenced any more except from search engines: redirect to /devices


jospoortvliet:
I like the tags, but let's not delete the Box page - I'd like to keep it as a piece of history, I suppose... As long as it is clear that the box is no longer available (and it is).

Could you revert the box deletion? I'll then merge it. I will also update the spreedbox link and move it to the available section, it is for sale... https://www.spreed.me/spreedbox/

jernst:
Done! /cc @jospoortvliet

PR #951, opened 2018-09-10
papercuts on athome

MariusBluem:
* remove a lost "."
* Appstore -> App Store
* link to page-yourdata.php instead of page-install.php to promote providers, hardware next to the classic download

Something for @shironextcloud and @jospoortvliet to review ;)


jospoortvliet:
Some good catches, Marius. I'm also hoping you will have some time to bring the translations back to 100% - kick me on IRC once they are so I can sync and deploy!

PR #952, opened 2018-09-11
update header, add flag

jospoortvliet:
* updated header. Works in firefox but not in Chrome - menu closes right away. Hoping @Espina2 knows...


jospoortvliet:
menu

What it looks like in Firefox (looks similar in Chrome, but the menu's on the right don't stay open so you can't click anything there)

Should also be tested on Safari, mobile & windows...

MariusBluem:
- **Devices-Icon** should be changed to 
https://thenounproject.com/jancborchardt/uploads/?i=638526
- sequence should be Download, Sign up, Devices and after that Enterprise (most people look at the last option first)
- **Demo** should be before **Get Nextcloud**
- instead of UN-flag we should use a **generic language icon** (
https://thenounproject.com/search/?q=Language&i=1824073)

Espina2:
@jospoortvliet fix it, by targeting Firefox and Safari and adding margin in that cases. This is fast fix, not the best one and should be revised when possible.

jospoortvliet:
@Espina2 you are MY HERO it works here now in chrome!

@Marius - the noun project wants my email and has all kinds of terms to agree with... do you have that icon in svg or so for me?

We can then perhaps also try to do something about the mail icon to, it is now the emoticon (...) of email.

The problem is that the bar on top is transparent and changes color. The text and icons should be white on the blue background but black on the white one. I haven't really figured out how to do that. Maybe @Espina2 has a hint, or you have an idea...

MariusBluem:
> @MariusBluem - the noun project wants my email and has all kinds of terms to agree with... do you have that icon in svg or so for me?

🏓ed @jancborchardt about this via Talk 😉 @jospoortvliet

Espina2:
To Change the svg icon color Just open it on gimp or inkscape. :p

jospoortvliet:
@Espina2 yeah, but I can't do that whenever a user visits the site ;-)

My problem is how to make the color of the svg change (or switch to another) when the bar changes color... But you add a class, I see, ```scrolled```, let me see if I can use that. Gimme some days :dagger:

Espina2:
I get it wrong them. This is quite easy to do. You just need the SVG without color. You already have a class on the header for that.

For example with basic scss: 

```scss
.headroom--not-top {
  .some-svg {
     fill: white;
     stroke: white;
  }
}
```

jospoortvliet:
@Espina2 yeah, but I can't do that whenever a user visits the site ;-)

My problem is how to make the color of the svg change (or switch to another) when the bar changes color... But you add a class, I see, ```scrolled```, let me see if I can use that. Gimme some days :D

Espina2:
@jospoortvliet 

Don't really get what you mean by "I can't do that whenever a user visits the site";

If the bar change color, it have a class toggling for that option, so you just need like the example i put on top describe which color the svg should have. 

You have no ideia on how to that, we can chat. ;)

jospoortvliet:
new-menu-front-page-scrolled
new-menu-front-page

Ok, above the new look. It is all SVG, it works in Firefox and Chrome, needs testing in safari and windows and other browsers I have no access too, though...

What does everyone think? Yes, Jan has already said we should move from these new svg flags to another set of svg flags but I couldn't figure out how to integrate those, these are easy... Just a span with a class. So for now, this is it, anyone should feel free to replace these flags with those from https://github.com/twitter/twemoji/tree/gh-pages/2/svg ;-)

But let's do that in a separate PR.

Oh, I disabled French, when the new header is there it might work but right now, https://nextcloud.com/fr_FR looks horrible due to the length of text.

PR #973, opened 2018-10-18
Improve simple signup

MorrisJobke:
* do not log error in case username is already in use and properly show this to the user
* show the overall time limit and advise to retry after this time
* allow 10 registrations within 5 minutes

bildschirmfoto 2018-10-18 um 09 59 49
bildschirmfoto 2018-10-18 um 10 10 12

### DIFF
--- a/lib/signup.php
+++ b/lib/signup.php
@@ -167,6 +167,15 @@ function request_account($request) {
 		if ($post['response']['code'] === 400 && $post['response']['message'] === 'invalid mail address') {
 			return new WP_Error('invalid_mail_address', 'invalid mail address', array('status' => 400));
 		}
+		if ($post['response']['code'] === 400 && $post['response']['message'] === 'Bad Request') {
+			$decodedBody = json_decode($post['body'], true);
+			if ($decodedBody !== null &&
+				isset($decodedBody['data']['message']) &&
+				$decodedBody['data']['message'] === 'user already exists') {
+
+				return new WP_Error('username_already_used', 'User is already existing', array('status' => 400));
+			}
+		}
 		error_log('Provider did not returned 201: ' . json_encode($post));
 		return new WP_Error('unknown_error', 'Something happened', array('status' => 400));
 	}
--- a/lib/signup.php
+++ b/lib/signup.php
@@ -111,8 +111,8 @@ function request_account($request) {
 	global $readerCountry;
 
 	$limit = [
-		'interval'     => 3660, // seconds
-		'num_requests' => 5, // number of requests allowed per interval
+		'interval'     => 300, // seconds
+		'num_requests' => 10, // number of requests allowed per interval
 		'user_ip'      => whatismyip() // getting the user IP.
 	];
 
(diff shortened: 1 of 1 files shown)


rullzer:
For now OK.

But the please retry is 5 minutes is not always right. Maybe after a minute they can retry again already. But better this way than the other way around :+1:

MorrisJobke:
> But the please retry is 5 minutes is not always right. Maybe after a minute they can retry again already. But better this way than the other way around 👍

I also thought about fetching the latest time. I could add this of course as well. Let me try a bit.

lib/signup.php
```diff
@@ -195,6 +191,10 @@ function request_account($request) {
 		return new WP_Error('rest_invalid_param', 'An unknown error occured', array('status' => 400));
 	}
 
+	$ttl  = $redis->ttl($rateId);
+	$redis->set($rateId, $rateLimit + 1);
+	$redis->expire($rateId, $ttl > 0 ? $ttl : $limit['interval']);
+
```
MorrisJobke: About this change I'm a bit skeptical: Before we raised the number once the request is sent out to the provider. With this change it is only counted once the request is checked for useful return values. That means "user existing" or "invalid email address" or "some 500/400 on the provider side" is not counted anymore.

This has the advantage, that a typo in the form (email incorrect, ...) is not counted anymore, but has the disadvantage of being able to brute force a bit more. :/ 

The more I think about this the less I want it to be included.

rullzer: What would be brute forceable here?
I'm fine with leaving it as it was for now btw if that is easier.

MorrisJobke:
> * allow 10 registrations within 5 minutes

Implication of this is change (maximum possible registrations): 

|      | 30 minutes | 1 hour | 8 hours |
| ------------- | ------------: |------------: |------------: |
| **Before**   | 5 users | 5 users | 40 users |
| **After** | 60 users  | 120 users | 960 users |

lib/signup.php
MorrisJobke: > What would be brute forceable here?

Already used email addresses. Let's kill this commit then.

rullzer:
I'm fine with that list for now. We should monitor it to see if it doesn't do :boom: of course

MorrisJobke:
Deployed.

skjnldsv:
> do not log error in case username is already in use and properly show this to the user

We removed it because this caused security issue iirc. 
I do not have any opinion on this, but this have been notified to me by some providers, stating that they were a bit thorn by this (seems like a bad things to do, but they were not that troubled by this either)

MorrisJobke:
> We removed it because this caused security issue iirc.
> I do not have any opinion on this, but this have been notified to me by some providers, stating that they were a bit thorn by this (seems like a bad things to do, but they were not that troubled by this either)

I just noticed in the logs, that people actually just re-submitted the form until they hit the rate limit, because it only said "Something happened" ... yes - thanks. We should be clear to the users and this is not really a problem, as it is rate limited on our side as well and they could do basically the same with the password reset form on the providers instance. ;)

skjnldsv:
Totally make sense :hugs: 
Thanks!

PR #982, opened 2018-10-25
update style of blog

jospoortvliet:
Soooo, I thought it'd be time to bring the style of our blog in line with the look we have on our newsletters these days ;-)

What do you think?
(fake content here, locally)
nextcloud_will_rock_fosdem_ _are_you_ready_for_it _too_ _nextcloud_-_2018-10-26_00 09 15
screenshot_20181026_000405
screenshot_20181026_000436
screenshot_20181026_000749
screenshot_20181026_000837

Besides ppl in review - input from @hebbet and everyone else more than welcome ;-)

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -10280,6 +10280,7 @@ li {
 body.changelog .content ul li {
   margin-top: 0;
   list-style: circle; }
+
 body.advisories .content ul li {
   margin-top: 0;
   list-style-type: initial;
--- a/assets/css/custom.scss
+++ b/assets/css/custom.scss
@@ -965,13 +965,16 @@ body.advisories .content ul li {
       transition: all 0.3s ease 0s; }
 
 /* A nice looking block quote */
+
 blockquote {
   background: none;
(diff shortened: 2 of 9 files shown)


jancborchardt:
Looking good! Only 2 things I can see so far:

- Is the white background box actually slightly blue? Should be white, with the slight transparency where the image is. (Blue might come from the background-background being blue, then we need another container with full non-transparent white.) But reading on blue looks silly – if the 2 container is too difficult I'd rather kill the slight transparency.
- Let's put a border-top-right-radius (and top-left-radius) of 10px on that! :) (we have the same on modals in Nextcloud 15)

jospoortvliet:
I'm extremely honoured that those are your only two comments, glad you like it!

Yeah, the blue is due to the background being blue. I will see if I can hack something, like a 2nd white below so it won't be blue, but I might have to disable the transparency.

Will add the rounded corners!

jospoortvliet:
Done. Required black magic but it works in chrome and firefox at least...

PR #986, opened 2018-10-29
Small text edits

shironextcloud:
edited the main part of the page. didn't touch the FAQ part, really don't like how defensive it is and how much we explain ourselves for doing something great.

### DIFF
--- a/page-include.php
+++ b/page-include.php
@@ -28,7 +28,7 @@
     <div class="container">
         <div class="row revealOnScroll">
             <div class="col-md-8 col-md-offset-2">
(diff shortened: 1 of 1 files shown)


jospoortvliet:
I guess we have to agree to disagree on that. I feel that we can't afford not to address it.

A vast majority of people in IT has heard little about the diversity debate, doesn't care much/spend much energy on it and has no axe to grind either way. What they HAVE heard or believe, esp if they're active in open source, is that we/our community/tech people judge each other and their work on technical skills, that we are a meritocracy and so on. You and I know that that is not really the case but nearly everyone thinks it is, if not explicitly and consciously, then in their subsconscious. They don't spend brain cells on this subject and just MIGHT have heard a few things, either from a feminist, troll, or both.

Not addressing that is self-defeating, I think. Yes, it is a bit defensive which I don't like much either and yes, it won't convince the trolls/bad people but then again, you don't fix those anyway. But most people are neither a feminist nor a troll. It's a matter of educating them. An alternative way of doing that is more than welcome, though.

PR #989, opened 2018-10-30
Use gender-neutral placeholder image

jancborchardt:
Instead of the obviously male-based "person.jpg" just a blank-style background (cropped from a full version of my photo so it’s the same background).

Because it looked pretty strange on the Include page especially. ;)
screenshot from 2018-10-30 14-07-55

Please review @jospoortvliet @shironextcloud – I did compress it locally, but this ImgBot probably has to go over it again because the previous file was much smaller still.


shironextcloud:
I agree, what do you think @jospoortvliet ?

jospoortvliet:
works for me :+1:

Issue #931, opened 2018-08-11
You have javascript disabled


splitt3r:
https://nextcloud.com/de/partners/ shows the "You have javascript disabled." message all the time. I tested it with FF and Chromium on Ubuntu. The message is shown with JS enabled.

eFrane:
Also true for Safari Mobile.

jospoortvliet:
Thanks, fixed!

t-markmann:
Please re-open: this seems to happen on every /de page. 
Have a look at https://nextcloud.com/de/files/ or https://nextcloud.com/de/federation/
Tested in Firefox & Chrome with disabled browser add-ons. Other pages without /de work fine :-)

PR #1002, opened 2018-11-19
Fix "You have javascript disabled" false positive

dorian-marchal:
:information_source: Easier to review with the [`?w=1`](https://github.com/nextcloud/nextcloud.com/pull/1002/files?w=1) flag.

---

Adds a missing `</section>` closing tag in page-clients.php.

This triggered "You have javascript disabled" false positive due to
`.alert--no-js` not being a direct child of `<body>` [0].

Note: this JS detection script should probably be more robust as this
is not the first time this issue occurs [1].

Also, adds two missing `</div>` tags and indents `<section>` children in
page-clients.php to make future missing closing tags more visible.

[0]: https://nextcloud.com/clients/
[1]: https://github.com/nextcloud/nextcloud.com/issues/931

### DIFF
--- a/page-clients.php
+++ b/page-clients.php
@@ -22,6 +22,7 @@
 		</div>
 	</div>
 </div>
+
 <section class="section--intro">
 	<div class="container">
 		<div class="row">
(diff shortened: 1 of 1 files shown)


MariusBluem:
> ℹ️ Easier to review with the ?w=1 flag.

Thanks 😄 ... Rest seems fine. 👍

PR #1005, opened 2018-11-20
Papercut about-page style

MariusBluem:
Before | After
------- | ----
bildschirmfoto 2018-11-20 um 15 15 33 | bildschirmfoto 2018-11-20 um 15 15 25

Signed-off-by: Marius Blüm <marius@nextcloud.com>


jospoortvliet:
Should be 'on your terms' but I'll fix that separately. Thanks!

PR #1006, opened 2018-11-22
12 is no longer supported!

jospoortvliet:
when oh when do we merge this :grin: 

@MorrisJobke I suggest in a week or two...

### DIFF
--- a/page-changelog.php
+++ b/page-changelog.php
@@ -19,7 +19,7 @@
     <ul>
     <li><a href="#latest14">Nextcloud 14</a></li>
     <li><a href="#latest13">Nextcloud 13</a></li>
-    <li><a href="#latest12">Nextcloud 12</a></li>
+    <li><a href="#latest12">Nextcloud 12 (unsupported!)</a></li>
     <li><a href="#latest11">Nextcloud 11 (unsupported!)</a></li>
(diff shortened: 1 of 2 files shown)


MariusBluem:
Added 8776162cf99d7740614fd9de2483cdd709ec243e to remove it from `page-security` too 😉

MorrisJobke:
> @MorrisJobke I suggest in a week or two...

Yep ... maybe also a bit longer. Early next year? Technically once 13.0.9 and 14.0.5 will be released, because then it didn't received the update.

jospoortvliet:
Well, if we find a security issue now, it won't be addressed, so I'd say there's risk already. Let's do it at release or shortly after 15 is out...

tflidd:
> Well, if we find a security issue now, it won't be addressed, so I'd say there's risk already. Let's do it at release or shortly after 15 is out...

In this case, you can release it already because it is not supported any more. And when someone reads it today, the planning for an upgrade starts today and not after the release of NC 15! I see no benefit in holding it back.

MorrisJobke:
> In this case, you can release it already because it is not supported any more. And when someone reads it today, the planning for an upgrade starts today and not after the release of NC 15! I see no benefit in holding it back.

True point 👍

PR #1008, opened 2018-11-23
Mobile Clients opens Desktop Clients

danielkesselberg:
Click on "mobile clients" opens "desktop clients" because jquery not ready on https://nextcloud.com/install/

### DIFF
--- a/page-install.php
+++ b/page-install.php
@@ -126,16 +126,15 @@
 </div>
 <?php require get_template_directory().'/install-instructions.php'; ?>
 <script>
-    $('#mobile').click(function () {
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Thank you. It works. To be honest, I don't even know how or why, I don't really know javascript :see_no_evil:

jospoortvliet:
Ok, so it didn't work - I reworked stuff, putting it in a install.js file loaded via require.js and that did the trick. Also cleaned up the other JS, putting it in there, too...

And while at it, added a stupid animation in the CSS. Because...

danielkesselberg:
Cool!

PR #1009, opened 2018-11-26
Install instructions: Arch Linux package available

michaellass:
nextcloud-client is now available in Arch's [community] repository, so there is no need to install from AUR anymore:
https://www.archlinux.org/packages/community/x86_64/nextcloud-client/

While here, change the distribution name. It calls itself "Arch Linux".


jospoortvliet:
Good news comes in the form of pull requests. Thanks!

PR #1030, opened 2018-12-25
add short business model explainer to signup page

jospoortvliet:
It was pointed out to me that we offer a 'free' service here, but don't really explain who pays and how. After all, our competition tends to monetize your private data to enable them to provide a service for free. I added a short explainer. Thoughts?

### DIFF
--- a/page-signup.php
+++ b/page-signup.php
@@ -111,7 +111,7 @@ $coreApps = ['files', 'calendar', 'contacts', 'spreed', 'mail', 'tasks', 'notes'
             <div class="text-center">
             <a href="<?php echo home_url('privacy') ?>" target="_blank" class="button button--blue button--arrow button--large"><?php echo $l->t('Our privacy policy'); ?></a>
             <p class="section--paragraph text-center"><?php echo $l->t('If there are any issues with your account, please note that we do NOT record any of your information on our side so we can not help you. Please contact the provider you signed up with.'); ?></p>
-            <p class="section--paragraph text-center"><?php echo $l->t('* free services are, of course, paid somewhere, by someone. Often with your data! Our providers all give a free 2-5GB storage shared account but offer larger storage and more options like a dedicated setup for a fee, which funds the free accounts. Consider paying for your account, it helps them provide this service!'); ?></p>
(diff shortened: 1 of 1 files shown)


includeyv:
I like this simple addition a lot. It is short indeed, and explains it all. Like I mentioned before, part of doing the right thing is mentioning what you are doing, and why you are doing it. This is especially important in (the curren) time wherein data-might-be-(mis)used-unexpectedly-awareness is not top of mind for many people. So, again, I like :-)

page-signup.php
```diff
@@ -111,6 +111,7 @@
             <div class="text-center">
             <a href="<?php echo home_url('privacy') ?>" target="_blank" class="button button--blue button--arrow button--large"><?php echo $l->t('Our privacy policy'); ?></a>
             <p class="section--paragraph text-center"><?php echo $l->t('If there are any issues with your account, please note that we do NOT record any of your information on our side so we can not help you. Please contact the provider you signed up with.'); ?></p>
+            <p class="section--paragraph text-center"><?php echo $l->t('* free services are, of course, paid somewhere, by someone. Often with your data! Our providers all give a free 2-5GB storage shared account but offer larger storage and more options like a dedicated setup for a fee, which funds the free accounts. Consider paying for your account, it helps them provide this service!'); ?></p>
```
jancborchardt: I’d simplify the wording here to:

> * free services are of course paid somewhere by someone. Our providers all give a free account with 2 to 5 GB storage but offer larger storage and more options like a dedicated setup for a fee, which funds the free accounts. Consider paying for your account, it helps them provide this service!

Using easier language and also leaving out the "Often with your data!" especially since people might think that at some of our providers you "pay with your data".

@jospoortvliet ok?

tflidd: > since people might think that at some of our providers you "pay with your data"

Hopefully not, but do you know for sure? Are these providers verified somehow or did they sign some sort of legal agreement? Therefore, it is perhaps a bit risky to talk about "our providers". We already had users complaining about certain providers for their terrible support and configurations that allow only a very limited range of features.

jospoortvliet:
Agreed, took Jan's modification.

PR #1036, opened 2019-01-10
data processing takes a bit

blizzz:
… but it should not run into timeouts.

It's already live^^

### DIFF
--- a/lib/signup.php
+++ b/lib/signup.php
@@ -284,7 +284,7 @@ function get_statistics($params) {
 			return [];
 		}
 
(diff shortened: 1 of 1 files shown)


lib/signup.php
```diff
@@ -284,6 +284,7 @@ function get_statistics($params) {
 			return [];
 		}
 
+        set_time_limit(0);
```
skjnldsv: Indentation is off :)

blizzz: Edited within Github… blame Microsoft :p

PR #1038, opened 2019-01-14
swap some "should"s for "must"s in Trademark policy

wiswedel:


jospoortvliet:
For english this is correct, I think it is clear enough.

PR #1042, opened 2019-01-15
Update teaser and banner images

jospoortvliet:


jancborchardt:
Nice! Some details:

- It would be good to keep "Mail" in the top bar. Instead of Circles for example, as no one knows the Circles icon, and "Mail" is more universally understood.
- Should we show the grid view? I mean it does look nice, but we don’t use it by default, and there will be some design changes too with it.
- Can we completely lose the blue on teaser.png please? ;)
- What are these big white dots here on top-banner@2x_mobile.png?
screenshot from 2019-01-16 16-37-10

jospoortvliet:
> * It would be good to keep "Mail" in the top bar. Instead of Circles for example, as no one knows the Circles icon, and "Mail" is more universally understood.

Ok, have to edit the image, but that's fine ofc

> * Should we show the grid view? I mean it does look nice, but we don’t use it by default, and there will be some design changes too with it.

It differentiates us... But we can check with Frank what he thinks.

> * Can we completely lose the blue on teaser.png please? ;)

Ok, what color to replace it with?

> * What are these big white dots here on [top-banner@2x_mobile.png](mailto:top-banner@2x_mobile.png)?

Ah, good catch. Will remove them.

jancborchardt:
> Ok, what color to replace it with?

Just transparency? Or white, if a background is needed. If we use any kind of background then it should be full Nextcloud blue like we do for our [Android app in the stores](https://play.google.com/store/apps/details?id=com.nextcloud.client), but not this light blue-ish. But again, for the website images no other color is needed – the screenshot content itself should stand out.

PR #1053, opened 2019-01-25
add search to our website

jospoortvliet:
Improved our 404 page and added a search page. I know it isn't super fantastic but it does the trick... There's a link in the footer.

screenshot_20190125_102752


jospoortvliet:
Going for it, it is better than we have (nothing) and we can improve later.

PR #1057, opened 2019-01-28
Separated mentos from coordinators and other edits

shironextcloud:


jospoortvliet:
tested, works, looks good.

PR #1062, opened 2019-01-30
Added form, removed mail buttons

shironextcloud:
Like for the includesubmit page, I don't really know what I'm doing here, did my best to do the basic stuff, please adapt as needed :) . Also, I don't know what you think of it, but I'd maybe put a blue background to that form section, and move the FAQ under the mentors? So we have two different blue sections. Also it makes more sense to have FAQ at the bottom. Didn't do that yet, up to you.


jospoortvliet:
I merged this in the other one ;-)

PR #1064, opened 2019-02-16
Change providers flags from uk to gb

silsha:
The icon set for the flags on https://nextcloud.com/providers/ has apparently changed.


jospoortvliet:
Great catch, thanks for fixing it!

PR #1081, opened 2019-03-29
Mail orders@nc.c and sent from sales

jospoortvliet:
Whaddayadunk?

### DIFF
--- a/page-ordersubmit.php
+++ b/page-ordersubmit.php
@@ -309,12 +309,12 @@ if(isset($_POST['email'])) {
 		$email_message .= "Signed terms: " . clean_string($terms) . "\n\n";
 
 		// create email headers
-		$headers = 'From: no-reply@nextcloud.com' . "\r\n" .
-			'Reply-To: ' . $email_from . "\r\n" .
+		$headers = 'From: sales@nextcloud.com' . "\r\n" .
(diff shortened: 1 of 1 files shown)


page-ordersubmit.php
```diff
@@ -309,12 +309,12 @@ function clean_string($string) {
 		$email_message .= "Signed terms: " . clean_string($terms) . "\n\n";
 
 		// create email headers
-		$headers = 'From: no-reply@nextcloud.com' . "\r\n" .
-			'Reply-To: ' . $email_from . "\r\n" .
+		$headers = 'From: sales@nextcloud.com' . "\r\n" .
+			'Reply-To: sales@nextcloud.com' . "\r\n" .
```
blizzz: @jospoortvliet :one thing to be sure: for the buyer it makes sense. for the Sales team, they now need to "Reply all" instead if hitting a normal reply. Will that work out?

jospoortvliet: Good point... How do I add both, if that is possible?

blizzz: @jospoortvliet comma separated should do it:
 `			'Reply-To: ' . $email_from . ",sales@nextcloud.com\r\n" .`

PR #1093, opened 2019-04-15
Improve wording of sub-headline on frontpage

jancborchardt:
The subheadline is a bit bumpy:

> Building self-hosted products that allow you to be productive without losing control

"products" and "productive" clash, and why the "building"? We should be talking about what the product does for people, not what we do?

E.g.:

> The self-hosted platform that makes you productive without losing control

What do you think @jospoortvliet @shironextcloud


shironextcloud:
I don't like "makes you productive", because it doesn't make you productive, it helps but there's no magic here. But tbh I don't have a better idea :/ Maybe @jospoortvliet has one?

PR #1113, opened 2019-05-15
Changelog for 14.0.11, 15.0.8 and 16.0.1

MorrisJobke:


MorrisJobke:
@jospoortvliet @marinelaa @shironextcloud The changelog script (https://github.com/nextcloud/github_helper#changelog-generator) now generates the full block also with the links to the download server and the date if it is executed on the day before the release (because I assumed, that's when we often prepare the PR here and thus it makes sense to assume it - otherwise those dates need to be adjusted).

jospoortvliet:
Nice work on that change, Morris!

jospoortvliet:
Tested, all good.

PR #1133, opened 2019-07-12
change tags for the Nextcloud Home/SME server

enoch85:
Figured that DYI doesn't fit this as it's fully assembled upon arrival.

cc @jospoortvliet


enoch85:
Thanks Jos!

PR #1137, opened 2019-07-20
Appdev page for on our website

jospoortvliet:
First draft... Still has some issues, including one section that I need to know more about - do we have ways to interact with external storage, workflows and other capabilities? What (else) is missing that I should point to?

appdevelopment – nextcloud 2019-07-20


jospoortvliet:
Implemented feedback ;-)
appdevelopment – nextcloud 2019-08-02

jancborchardt:
Looking at the page again and again there’s way too many buttons. 5 buttons below each other needs to be a list instead. 1 button can be primary, if there’s a main action. But like this there’s too much.

jospoortvliet:
appdevelopment – nextcloud 2019-08-07
Contribute – nextcloud 2019-08-07
after more feedback ;-)

jancborchardt:
Just small feedback like:
- the subline in the header should be better readable (regular font)
- the "Everything you need" line can be cut
- screenshots need to be clearer, or we just use icons

So we can also merge first and polish that later :)

jospoortvliet:
Approved by the almighty @jancborchardt and Frankie

PR #1141, opened 2019-08-02
Update home and about page page

jospoortvliet:
* Put our products and customers more prominent on home page
* add a scroll down animation instead of the button
* clean up about page a little
* add our mission and principles to about page
Previews:
about – nextcloud 2019-08-02
nextcloud 2019-08-02

### DIFF
--- a/assets/css/pages/about.css
+++ b/assets/css/pages/about.css
@@ -56,42 +56,6 @@ section {
     section .newsrow .col-md-4 {
       padding: 15px; }
 
-.section--products {
-  margin: 80px 0 0 0;
-  padding: 90px 0 70px 0;
-  background-color: #eee; }
-  .section--products .row {
-    margin-bottom: 10px; }
-  .section--products img {
-    padding: 20px; }
-  .section--products .description {
-    min-height: 100px;
-    margin: 0 20px; }
-    .section--products .description .title {
-      display: block;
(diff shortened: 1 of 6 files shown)


jancborchardt:
Very cool! I would only change 1 thing on the front page: Put the files/talk/groupware first, and the "Powering collaboration" with blue background, and monochrome logos. Cause grey background looks a bit bland.

jospoortvliet:
@jancborchardt hmmm, tried that, but it doesn't feel great to me:
Screenshot_20190805_094155
Filter is: `filter: invert(1) saturate(0%) contrast(.35) brightness(130%);`

(If you want to experiment on the current home page to find better settings - add class `quote` to section `section--customers` to make it blue and then filter the images)

One idea I had (copied from somebody else of course):
Screenshot_20190805_095837

jancborchardt:
@jospoortvliet sorry, I should have been specific. With logos in monochrome I meant that they should be a single color – white in this case – not greyscale. This would work very well on the blue.

See e.g. on https://about.gitlab.com where the customer logos are single color too, in their case grey.

jancborchardt:
So so so so so much better! Great job @jospoortvliet! :)

jospoortvliet:
Thanks, @jancborchardt - it's your idea, I just do some css and html :smile:

PR #1148, opened 2019-08-15
Add audioplayer.svg

skjnldsv:
Add the audioplayer app

@jospoortvliet lots of css are changed but not properly compiled into their own css. Is this an issue? Also, any reason our css files are not compiled as compacted files? I think we'll gain a lot of speed for that :)


jospoortvliet:
1. wrt files that aren't compiled - the modules are not, as they are just  included in custom.css
2. wrt compaction - yeah, that'd be nice, but we have nothing set up to do that automatically, we deploy with a git pull remember  ;-)
Something for a new website...

PR #1157, opened 2019-09-10
pricing page: fixes and enhancements

wiswedel:
- wording fix
- better visibility for optional components

### DIFF
--- a/page-pricing.php
+++ b/page-pricing.php
@@ -76,7 +76,7 @@ installation call optional)');?>" rel="tooltip"><?php echo $l->t('Installation a
 				<h1 class="">Optional components</h1>
 				<div class="line"></div>
 				<div class="list">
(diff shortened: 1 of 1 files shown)


jospoortvliet:
looking good and works. But if we add these, shouldn't we also add the outlook add-in (for all) and GS (for Premium)? And why is collabora available for basic, but not ONLYOFFICE?

wiswedel:
@jospoortvliet as for GS, I agree, let's add that.
The other ones should be correct already 🤔
Do you have a screenshot?

jospoortvliet:
Screenshot_20190910_151137

wiswedel:
> Screenshot_20190910_151137

fixed by https://github.com/nextcloud/nextcloud.com/pull/1157/commits/b4c8119db7661a72f697696381ddeef778a5ea99

PR #1160, opened 2019-09-19
Updated UBOSbox text to be consistent with what we have on our site

jernst:


jospoortvliet:
Sorry, that I only see it this late!

PR #1165, opened 2019-10-02
Should fix the issue with dev documentation

LaPalice:
I hope it doesn’t break other stuff 🙈


LaPalice:
@jospoortvliet Github wont let me add reviewers so I ping you here

PR #1173, opened 2019-10-12
Correct the bug bounty amount

xh3n1:
The highest possible reward is: $10,000: https://hackerone.com/nextcloud


jospoortvliet:
Very good catch, thanks!!!

PR #1181, opened 2019-10-26
Revealonscroll

jospoortvliet:
removing the revealonscroll thing - @jancborchardt wanted it and honestly, so did I ;-)

### DIFF
--- a/assets/css/custom.scss
+++ b/assets/css/custom.scss
@@ -904,7 +904,7 @@ i[class*='fa-'] {
 }
 
 .revealOnScroll {
-  opacity: 0;
+  opacity: 1;
 }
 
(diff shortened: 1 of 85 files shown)


assets/css/custom.scss
```diff
@@ -904,7 +904,7 @@ i[class*='fa-'] {
 }
 
 .revealOnScroll {
-  opacity: 0;
+  opacity: 1;
```
jancborchardt: I would rather remove this completely, as otherwise this overwrites other values which might have been set. :)

jospoortvliet: Yeah, true, I've now removed this anyhow... This could get cleaned, too. Will do separately.

PR #1186, opened 2019-11-06
Fix Simple Sign up layout on mobile

jancborchardt:
Fixes the overlap cc @jospoortvliet @skjnldsv. Total hack but well

@skjnldsv @jospoortvliet can we get rid of the `revealOnLoad`? Also there’s a horizontal scrollbar and content is squished to the right?

## Before
simple signup before

## After
simple signup fixed

### DIFF
--- a/assets/css/pages/signup.scss
+++ b/assets/css/pages/signup.scss
@@ -66,7 +66,7 @@ section {
 
 @media (max-width: 768px) {
     #register {
-        margin-top: -500px;
+        margin-top: -380px;
     }
     .background .topheader {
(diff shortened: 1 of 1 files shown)


jospoortvliet:
I dunno where the horizontal scrollbar or revealOnLoad are, but I guess in the JS part...

jancborchardt:
> I dunno where the horizontal scrollbar or revealOnLoad are, but I guess in the JS part...

Yup, follow-up in https://github.com/nextcloud/nextcloud-register/pull/234 :)

PR #1199, opened 2019-11-28
update Debian/Ubuntu repository link

duub:
As mentioned in this issue, the main repository is the updated link: https://github.com/nextcloud/desktop/pull/1629#issuecomment-558937971


duub:
I've been using the current link with several computers and now I've to replace it with this other one because the current source stopped to update to new versions. And there's no reference to anywhere...

PR #1205, opened 2019-12-11
privacy policy update if we decide to use podlove

jospoortvliet:


tflidd:
Is there no other way to implement the subscribe button? If you subscribe for a service it is clear that you send data to them. But a subscribe button also affects all other users just visiting the website (and you load/send data from/to additional servers).

jospoortvliet:
Yeah, I'll have a look during 36C3 if we can avoid loading it on all pages... I wish I had time to move the site to Hugo to avoid that :dagger: :bomb: :keyboard:

jospoortvliet:
So we don't use the podlove add-in, but we do use the podlove player on nextcloud.com/podcast...

PR #1255, opened 2020-03-12
Prevent self-xss with burpsuite

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


jospoortvliet:
I don't understand it but I've deployed it :D

PR #1275, opened 2020-04-09
Update some node packages to make npm audit happier

ChristophWurst:
This resolves a few of the npm security audit alerts. Some are still to do.

The actual changes are in package.json. But since someone thought adding node_modules to git was a good idea I'm giving you all of that :)

I did not test this change. So please do.


gary-kim:
> The actual changes are in package.json. But since someone thought adding node_modules to git was a good idea I'm giving you all of that :)

Wait what? WHY? :cry: 

I also found this amazing PR by ImgBot that optimized images in `node_modules` so the hashes aren't even correct anymore... #1174

jospoortvliet:
> Wait what? WHY?

because our website is screwed up and it has to be 100% contained in the repo. We deploy with a 'git pull', no other stuff, so an update to NPM packages would never happen.

There's a reason we wanted to rewrite the website to Hugo ;-)

jospoortvliet:
meanwhile, I'm sad to have to say the update breaks our website. The header and front page animations stop working. I didn't write nor understand any of that code so I'm not even considering debugging it, sorry.

Issue #1278, opened 2020-04-25
changelog for 18.0.4


r2evans:
18.0.4 was released yesterday on the blog and on hub.docker.com, but https://nextcloud.com/changelog/ still shows only 18.0.3. Thank you!

r2evans:
Sure, but from 5 seconds ago ...

![image](https://user-images.githubusercontent.com/5815808/80289633-a5cc6700-86f4-11ea-8b7d-bf55ac9c36f2.png)

kesselb:
I guess someone has to deploy the changelog update. In the meantime check https://github.com/nextcloud/nextcloud.com/blob/master/page-changelog.php for the changes.

donvipre:
@kesselb: Do you know who could do this? To be honest, the source code philosophy at the nextcloud.com repository appears a bit unclear to me.

kesselb:
I guess @jospoortvliet or @nextcloud/infrastructure

r2evans:
Looks like it has since been updated, thanks.

PR #1279, opened 2020-04-25
Fix: correctly name latest version to 18.0.4

donvipre:
something went wrong with the latest changelog update.

Fixes: #1278


kesselb:
cc @jospoortvliet

PR #1280, opened 2020-04-30
sorted teams a bit better

wiswedel:
Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>

### DIFF
--- a/page-team.php
+++ b/page-team.php
@@ -63,6 +63,18 @@
 				<p class="teamdescription">Andreas studied Economics and has 20 years experience in Business Development, Partner- and Sales Management. Andreas worked for companies like Oracle, EMC, Yahoo and Google and also for start ups in the Cloud Computing and Online Marketing space.</p>
 			</div>
 		</div>
+	<div class="col-md-4">
+			<div class="teamp">
+				<img class="img-responsive img-circle teamimg" alt="photo" src="<?php bloginfo('template_directory'); ?>/assets/img/people/markus.jpg" />
+				<h2 class="teamname">Markus Windecker</h2>
+				<div class="teammeta">
+					<h3 class="teamtitle">Account and Partner Manager</h3>
+					<h4 class="teamlocation">lives in Herrenberg, Germany</h4>
+					<p class="teamsocial"><a href="http://www.linkedin.com/in/markus-windecker-43a723b2"><i class="fa-linkedin fa"></i></a></p>
+				</div>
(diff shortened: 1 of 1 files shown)


wiswedel:
@ecblack Could you please add some info about you on the Team page? I already added you but you need to fill in some text. Just checkout this branch and commit to it.

wiswedel:
by the way: Is Patrick as "Sales Manager" still correct?

jospoortvliet:
> by the way: Is Patrick as "Sales Manager" still correct?

mja, maybe not, but I believe he didn't want to be something like accounting. So let's keep it. Thanks for the PR, will check and merge now.

jospoortvliet:
also, what logic did you use to sort 'teams'? Before, I had senior management first, then sales, then marketing, then engineering, each of the alphabetically by first name... now it seems random, I mean, Markus Windecker is the 3rd person? I'll put the ordering back to something sane if you don't mind.

jospoortvliet:
Was good to update things... So what I did now is:
* created a management team, with Frank first and the rest alphabetically.
* Created a sales team, alphabetically.
* Created a marketing team, alphabetically.
* engineering & support - also alphabetically.

We still have to add Nils, Kevin and Julien, so let's wait with merging this until they are added and then I'll put it live. We are at 42 team members, by the way 🛣 📖

wiswedel:
Thanks for adding team headlines, @jospoortvliet. That's in the end what I intended :-)
If you don't mind, I'd like Sales Engineering to be in a separate section so it doesn't get mixed up with the AMs.

jospoortvliet:
> Thanks for adding team headlines, @jospoortvliet. That's in the end what I intended :-)
> If you don't mind, I'd like Sales Engineering to be in a separate section so it doesn't get mixed up with the AMs.

are you sure? It seems a bit detailed - you also have ppl handling inbox etc... I can do it, of course, but if it says 'sales engineer' that seems to do the trick and I don't know what problem you exactly want to prevent?

wiswedel:
Pre-Sales is a different profession than Account Management, even though of course both work closely together. It's important for the SE's trustworthiness towards the customer to maintain a strict boundary between Pre-Sales and Sales in the process.

So long story short, it's kind of a matter of the SE profession's self-conception.

jospoortvliet:
> Pre-Sales is a different profession than Account Management, even though of course both work closely together. It's important for the SE's trustworthiness towards the customer to maintain a strict boundary between Pre-Sales and Sales in the process.
> 
> So long story short, it's kind of a matter of the SE profession's self-conception.

allright, I'll separate it!

PR #1288, opened 2020-05-27
Add some padding for complaint section

kesselb:
|Master|This PR|
|-|-|
|Screenshot from 2020-05-27 23-03-10|Screenshot from 2020-05-27 23-04-34|

I'm not able to test it locally so ;) Please double check if everything else still works. Just moved the div around with firefox :see_no_evil:

### DIFF
--- a/compliance.php
+++ b/compliance.php
@@ -4,7 +4,6 @@
 <section class="section--compliance">
 <div class="container-fluid banner quote">
     <div class="container">
-    <div class="row">
         <div class="col-md-6">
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Mja, removing the row brings other issues. The problem is that we have a removal of horizontal padding in the .quote class, which was meant to only remove vertical padding. I've fixed that now with https://github.com/nextcloud/nextcloud.com/commit/6d87d19d2a132a1589c5b69a6748537d825f4e4b

-> close. Thanks!

PR #1309, opened 2020-06-26
move the Home/SME server to top

enoch85:
It should have been like that from start, just didn't know how to fix it back then.

Please review @jospoortvliet @tflidd

### DIFF
--- a/page-devices.php
+++ b/page-devices.php
@@ -54,23 +54,6 @@ require(["require.config"], function() {
         </div>
         <div class="row feature--block">
             <div class="col-md-5">
-                <a href="https://ownyourbits.com/nextcloudpi/"><img src="<?php bloginfo('template_directory'); ?>/assets/img/box/nextcloudpi.png" class="img-responsive featureimg" /></a>
-            </div>
-            <div class="col-md-7">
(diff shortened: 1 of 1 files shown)


tflidd:
I wouldn't place my solution first. So what is our criteria for ordering, the most popular, the oldest, ....?

enoch85:
> the most popular, the oldest, ....?

The newest are placed on top. NetcloudPi was founded before the server. Hence the change.

jospoortvliet:
Pfff. The order was totally random, I guess I did add stuff when I became aware of it from bottom to top but I'm not sure. If anything, alphabetical naming probably makes most sense. Don't you agree?

Also, Turrix Mox is now for sale for a while already and I should move the Spreedbox down to 'no longer available'. That I can do in a separate PR.

enoch85:
@jospoortvliet So in other words, 

1. Remove NextcloudPi
2. Merge this PR?

jospoortvliet:
Yeah, let's go for one. NextcloudPi is super cool, but NOT a device. So it doesn't belong there. Modify your PR to remove it, pls, then I'll merge.

enoch85:
@jospoortvliet Done. :)

PR #1310, opened 2020-06-27
Update page-collaboraonline.php

gorn:
We need to be more specific about limitations of the Collabora Development version. Limitation to 10 documents is only mentioned after the fact at the moment.

### DIFF
--- a/page-collaboraonline.php
+++ b/page-collaboraonline.php
@@ -100,7 +100,7 @@
 					<iframe width="100%" height="315" src="https://www.youtube-nocookie.com/embed/8iuSFKVl-xQ" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>
 				</div>
 			</div>
-			<p class="section--paragraph"><?php echo $l->t('Collabora Online Development Edition (CODE) is free and under heavy development, adding features and improvements all the time! Enterprise users have access to the more stable, scalable Collabora Online Enterprise version through a Nextcloud support subscription.');?></p>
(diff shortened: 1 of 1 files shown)


tflidd:
If the number of users is limited, the exact number would be interesting as well.

Can only 10 users open a document at the same time or is it really the number of documents. So you could have 2 documents that are opened by 100 users each?

👍 for adding this to the documentation/website

jospoortvliet:
Correct me if I'm wrong, but I don't believe there is a hard limit built in, just a warning like we do in Talk. So there's no need to tell people the number like it is so black ad white - as it depends on the hardware you throw at it...

gary-kim:
> Correct me if I'm wrong, but I don't believe there is a hard limit built in, just a warning like we do in Talk. So there's no need to tell people the number like it is so black and white - as it depends on the hardware you throw at it... On a Raspberry Pi (where it currently doesn't run, I know) it would be maybe 5, and on a Threadripper it might be 50 ;-)

It is a hard limitation: https://git.libreoffice.org/online/+/refs/heads/distro/collabora/code-4.2.0-4/configure.ac#536

The exact limitation is 20 connections for 10 different documents.

It's not particularly hard to change the configuration and compile your own version but the compile can take several hours and is quite confusing if you're not used to doing that sort of thing.

jospoortvliet:
Ok, a hard limitation - that we should tell users about. Thanks, will merge!

jospoortvliet:
@gary-kim thanks again for going through the trouble of this and sorry for the frustration!

kiki78:
Just for information, I use Collabora Code with NextCloud since some years and now, with last version (4.2.4 in Docker) I receive the limitation message for each document opening, although there is only 1 user and 1 document opened (verified in Collabora web admin console).
Is this normal ?

jospoortvliet:
> Just for information, I use Collabora Code with NextCloud since some years and now, with last version (4.2.4 in Docker) I receive the limitation message for each document opening, although there is only 1 user and 1 document opened (verified in Collabora web admin console).
> Is this normal ?

This must be a bug/issue. Try to report it to them, I would say.

PR #1313, opened 2020-07-02
Revert "No more freeplans by owncube :("

nickvergessen:
Reverts nextcloud/nextcloud.com#1308

### DIFF
--- a/assets/providers.json
+++ b/assets/providers.json
@@ -41,6 +41,20 @@
          "imagename":"hostiso.png"
       },
 	{
+         "title":"OwnCube e.U.",
+         "url":"https:\/\/owncube.com",
+         "flags":[
+            "de",
+            "all"
+         ],
+         "freeplans":true,
+		 "supports":"both",
+         "specializes":[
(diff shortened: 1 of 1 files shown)


assets/providers.json
```diff
@@ -39,6 +39,20 @@
...
+      },
+	{
+         "title":"OwnCube e.U.",
+         "url":"https:\/\/owncube.com",
+         "flags":[
+            "de",
+            "all"
+         ],
+         "freeplans":true,
+		 "supports":"both",
```
nickvergessen, line 51:
```suggestion
         "freeplans":false,
         "supports":"both",
```

kesselb:
Fine by me but it seems the file is unused anyway :)

nickvergessen:
No, it's the partners page. Just not the preferred providers

jospoortvliet:
Ok, I should just remove this file, as I don't think it does anything anymore. The actual data for this is only on the server and has to be manually edited. It isn't in github...

jospoortvliet:
FYI this list is now here: https://github.com/nextcloud/providers

jospoortvliet:
I will ssh into the server and removing ownCube.

PR #1328, opened 2020-07-31
Remove [stable1X] from changelog

MorrisJobke:
This basically only removes the not needed tags we use in our pull requests.

I fixed this some weeks ago already in the changelog script: https://github.com/nextcloud/github_helper/commit/81b8f1233f164ae5d2c9d05e1fcc41309120604c#diff-6b306396aa9fdbc560492611d41dd56f

So a `git pull` in the `github_helper` directory will update the script and then the generated changelog will not contain those entries anymore. @marinelaa You are adding the changelogs, right? If you need help or something doesn't work, just ping me.

cc @jospoortvliet


jospoortvliet:
Tested, no problems, put live.

PR #1347, opened 2020-09-08
link to try.nc.c and make get Nextcloud a button

jospoortvliet:
Screenshot_20200908_200734
Screenshot_20200908_200804


ma12-co (changes requested):
I would say that the new button should have a height of 32 px at least. Ideally within the apps we're trying to bring all the buttons to 44px height. That might be a bit too much for the header though

jospoortvliet:
> I would say that the new button should have a height of 32 px at least. Ideally within the apps we're trying to bring all the buttons to 44px height. That might be a bit too much for the header though

yeah, it'd be lovely, but this is actually the default size of the 'small' button, I'm already happy it fits... Making it higher makes the menu overlap it (and I'm not sure I can fix that). But let me give it a try.

jospoortvliet:
So, I managed to fix the overlap with z-index, and move it all in place, with the normal button size:
Screenshot_20200909_071145
there is a not-awesome mouse-over issue when you come from the left menu, but it doesn't happen on direct mouse-over:
1280p15-magick-gifsicle

What do you think?

ma12-co:
Can we just link to the install page and remove the popover completely for this button?
This menu makes things easy for a somebody that already knows Nextcloud, but the install page is better for first time users as things can be explained much better.

jospoortvliet:
> Can we just link to the install page and remove the popover completely for this button?
> This menu makes things easy for a somebody that already knows Nextcloud, but the install page is better for first time users as things can be explained much better.
> We can also work on making the info laid out in the install page super clear in a follow-up

we can, but I would suggest we do that as part of improving the install page then. So keep this as-is and we can do a change to the install page which then removes the menu. My main motivation is that while these were reasonably quick changes for the install page - well, my past experience with it hasn't included the term 'fast'.

jancborchardt:
For accessibility, we need to make sure there’s hover as well as focus feedback on the "Get Nextcloud" button. So e.g. making the background blue while keeping a white border & text.

jospoortvliet:
> For accessibility, we need to make sure there’s hover as well as focus feedback on the "Get Nextcloud" button. So e.g. making the background blue while keeping a white border & text.

yeah, it's supposed to do that, but somehow it doesn't. let me fiddle with it some more.

jospoortvliet:
Screenshot_20200909_174823
Screenshot_20200909_174853

So I had to move the menu's down a bit, to not overlap the button. That means the gap between the text and the menu is a bit bigger than before  but I guess/hope it isn't a disaster.

jospoortvliet:
sheesus. That did NOT go well.

jospoortvliet:
Anyway, fixed it. @ma12-co can you test and let me know if you see issues? If not, let's squash-merge this...

ma12-co (approved):

jospoortvliet:
waiting for https://github.com/nextcloud/nextcloud.com/pull/1347

PR #1348, opened 2020-09-09
Whitespace

jospoortvliet:
This has a lot of whitespace improvements - but pure consistency will require somebody going over the various pages and fixing things manually I'm afraid :(

I did also add the card thing on the front page:
Screenshot_20200909_082524

Will add the same on the other pages where I have a similar thing.

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -3385,52 +3385,75 @@ tbody.collapse.in {
     .container-widest .container-widest {
       width: 1770px; } }
 
+section {
+  padding-top: 100px;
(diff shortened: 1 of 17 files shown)


jospoortvliet:
now also put this on the hub page:
Screenshot_20200909_083334

jancborchardt:
For accessibility, we should make sure that we don’t have generic buttons saying "Learn more" – the button text should always be relevant to the topic, e.g. "More about Nextcloud Hub".

jospoortvliet:
Ok, can make that change ;-)

PR #1351, opened 2020-09-09
rename front image to avoid add blockers

jospoortvliet:


jancborchardt (approved):
Not tested but should work against adblocking (with uBlock it already works).

PR #1366, opened 2020-10-04
Update page-pr20201003-6.php

splitt3r:
@jospoortvliet there was sth. wrong. Was intended this way?

### DIFF
--- a/page-pr20201003-6.php
+++ b/page-pr20201003-6.php
@@ -44,7 +44,7 @@
 <p>Nextcloud GmbH<br />
 Jos Poortvliet<br />
 Phone: +49 (0) 171 121 7528<br />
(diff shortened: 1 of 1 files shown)


page-pr20201003-6.php
```diff
@@ -44,7 +44,7 @@
 <p>Nextcloud GmbH<br />
 Jos Poortvliet<br />
 Phone: +49 (0) 171 121 7528<br />
```
splitt3r, line 46: Maybe `+49 (171) 121 7528` just like the other two numbers here?

jospoortvliet (approved):
Thanks!!! Good catch. I fixed it in a bunch of them, but missed this one. The markdown annoyingly generates this instead of a normal link...

page-pr20201003-6.php
jospoortvliet: mwah, details details... If you want, sure, but not a big deal.

PR #1378, opened 2020-10-23
Feature/noid/improve install page

ma12-co:


ma12-co:
Jos could you please check the simple signup overlay? The url seems good but it doesn't work on my local environment, returns 404

jospoortvliet:
> Jos could you please check the simple signup overlay? The url seems good but it doesn't work on my local environment, returns 404

You have to add the page in wordpress ;-) Then it will work. All good here.

jospoortvliet:
So it all looks good to me... Do you want to merge this first, or first also do the popups?

(I would suggest to merge this but your call)

jospoortvliet:
So feedback from @camilasan on the design: She didn't realize at first it was an install page. The other one had big and clear, with icons: server/desktop/mobile. here is a lot more text you have to read

I do think the icons might be nice to have. We have to explain the server vs client vs etc etc, perhaps a graphic/illustration can be used?

ma12-co:
Yes I also thought that we need some kind of illustration, but since the design was previously approved I thought that this was good enough to get in. @jancborchardt do you have any suggestions here? 
We could go with: 
1) Server - server icon
2) Clients - Mobile icon
3) Apps - **suggestions?**

jospoortvliet:
For review, this is what it looks like.
screencapture-127-0-0-1-wordpress-install-2020-11-09-13_39_09
screencapture-nextcloud-install-2020-11-09-13_39_41

jancborchardt:
Hmm, so looking at the screenshots, the previous version seems to work a bit better, no?
- It has less text and is thus clearer
- The core things are more visible directly, especially on desktop
- The 1-2-3 is not really so visible, and doesn’t look like a process since 
- Not sure after all if the install page is the right place to promote the different apps. Cause a lot are now installed with Hub. What we could do is have that show post-download, so it’s really incremental?
- Visually looks better cause there are no long text lines
- Has icons to quickly communicate the use of the specific section

A specific issue which I saw previously when watching people use the install page is that e.g. they would download the server package to their desktop and think this would be the Nextcloud app. This is one of the core issues we need to wrangle cause the rest is exactly the same as with other proprietary platforms.
A possible solution there would be for example to focus on the Mobile & desktop apps with perhaps 2 columns, and then put the server _below_. Cause people who know about self-hosting will see that all over, and can see it there, but also in the setup dialogs of the mobile and desktop clients. In the download/install section we could focus on mobile & desktop thus.

I still think the main improvement we need to make on the install page is make it more direct and super obvious – and that needs some type of imagery: screenshots, illustrations or icons. E.g. look at https://signal.org/download/, so dead-simple and straightforward, and it takes into account the system you are accessing it from.

jancborchardt:
Regarding the icons, what about those:

1. Server: https://github.com/nextcloud/nextcloud.com/blob/master/assets/img/icons/secure-server.svg
2. Desktop & mobile: https://github.com/nextcloud/nextcloud.com/blob/master/assets/img/icons/mobile.svg and laptop part of https://github.com/nextcloud/nextcloud.com/blob/master/assets/img/icons/tools.svg
3. App ecosystem: https://github.com/nextcloud/nextcloud.com/blob/master/assets/img/icons/check-lock.svg

jospoortvliet:
Ok, so let's go with those icons. Then for the texts, how about these:

With Nextcloud you control where your documents, photos and chat messages are! Choose one of our recommended providers or download our server software and run it where you want.

The second is good as it is.

You can extend the functionality of Nextcloud, and connect a wide variety of mobile and third-party apps to Nextcloud.

page-install.php
```diff
@@ -10,35 +10,31 @@
...
-		<p><?php echo $l->t('There are several ways to get your own Nextcloud for you and your data.');?></p>
-        <p class="text-center"><a class="button button--blue button--large" href="#instructions-server"  ><?php echo $l->t('Download for server');?></a></p>
-        <p class="text-center"><a class="button button--white" href="<?php echo home_url('signup'); ?>"><?php echo $l->t('Sign up with a provider');?></a></p>
-        <p class="text-center"><a class="hyperlink" href="<?php echo home_url('devices') ?>"><?php echo $l->t('Buy devices');?></a> &nbsp; &nbsp; <a class="hyperlink" href="<?php echo home_url('providers') ?>"><?php echo $l->t('Find a provider');?></a> <br> <a class="hyperlink" href="<?php echo home_url('enterprise') ?>"><?php echo $l->t('Enterprise Solutions');?></a></p>
+		<h3><?php echo $l->t('Get your Nextcloud server up and running');?></h2>
+		<p><?php echo $l->t('At Nextcloud, we provide the software and <strong>you decide what to do with it.</strong> In order to use Nextcloud, you first need to install the Nextcloud server. You can either install it on a home server, on a cloud provider of your choice, or  use one of our erecommended providers. ');?></p>
```
jospoortvliet, line 18:
```suggestion
		<p><?php echo $l->t('With Nextcloud you control where your documents, photos and chat messages are! Choose one of our recommended providers or download our server software and run it where you want.');?></p>
```

page-install.php
```diff
@@ -10,35 +10,31 @@
...
+	<div class="install-section blue-border">
+		<div class="numbadge install-page__number">
+			<p>1</p>
 		</div>
-		<h2><?php echo $l->t('Server');?></h2>
-		<p><?php echo $l->t('There are several ways to get your own Nextcloud for you and your data.');?></p>
-        <p class="text-center"><a class="button button--blue button--large" href="#instructions-server"  ><?php echo $l->t('Download for server');?></a></p>
-        <p class="text-center"><a class="button button--white" href="<?php echo home_url('signup'); ?>"><?php echo $l->t('Sign up with a provider');?></a></p>
-        <p class="text-center"><a class="hyperlink" href="<?php echo home_url('devices') ?>"><?php echo $l->t('Buy devices');?></a> &nbsp; &nbsp; <a class="hyperlink" href="<?php echo home_url('providers') ?>"><?php echo $l->t('Find a provider');?></a> <br> <a class="hyperlink" href="<?php echo home_url('enterprise') ?>"><?php echo $l->t('Enterprise Solutions');?></a></p>
+		<h3><?php echo $l->t('Get your Nextcloud server up and running');?></h2>
```
jospoortvliet, line 17:
```suggestion
		<h3><?php echo $l->t('Get your Nextcloud server up and running');?></h3>
```

page-install.php
```diff
@@ -10,35 +10,31 @@
...
-			<i class="fa-mobile"></i>
+	<div class="install-section install-section--last">
+		<div class="numbadge install-page__number">
+			<p>3</p>
 		</div>
-		<h2><?php echo $l->t('Mobile');?></h2>
-		<p><?php echo $l->t('Use your Nextcloud on the go with our Android and iOS apps.');?></p>
-        <p class="text-center"><a class="button button--blue button--large" href="#install-clients" rel="tooltip" id="mobile" data-toggle="popover" title="Mobile Clients"><?php echo $l->t('Mobile apps');?></a></p>
+		<h3><?php echo $l->t('Explore the app universe');?></h2>
+		<p><?php echo $l->t('Nextcloud counts with a large and growing number of applications. Many of them offer clients  for mobile, both offically supported and developed by our community.');?></p>
```
jospoortvliet, line 35:
```suggestion
		<p><?php echo $l->t('You can extend the functionality of Nextcloud, and connect a wide variety of mobile and third-party apps to Nextcloud.');?></p>
```

ma12-co:
closing in favor of #1397

PR #1424, opened 2021-01-19
update design of devices page

jospoortvliet:
I wanted to make the devices page a bit more compact, as there will probably be more and more over time and getting an overview gets harder. Of course, on mobile this doesn't make it much better but on a bigger screen it helps.

## Before
screencapture-nextcloud-devices-2021-01-19-13_59_53

## After
screencapture-127-0-0-1-wordpress-devices-2021-01-19-14_00_05

### DIFF
--- a/assets/css/pages/devices.css
+++ b/assets/css/pages/devices.css
@@ -42,11 +42,28 @@ section {
   section .section--paragraph__title {
     padding: 0; }
 
-.feature--block {
+.section--devices .feature--block {
   padding-top: 60px; }
-
-.row {
+.section--devices .row {
   padding-top: 60px; }
+.section--devices .col-md-4 {
+  padding: 20px; }
+.section--devices .block-container {
+  height: 800px;
+  overflow-y: auto;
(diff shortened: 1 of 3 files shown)


jancborchardt:
Yeah, better to put it in a grid view. Some enhancements:
- Instead of using the "cards" look, just put some more whitespace inbetween
- Scrolling inside the page is a big no-no ;) Let’s make sure the texts are all roughly the same (short) length?
- The images should have the same height, so that the product title is at same height and we can get all the entries roughly same height.
- Nicer images for the UBOSboxes (no table background) and the Turris Mox (no Nextcloud logo or fade) would look nice
- The buttons should have different and relevant labels instead of "Learn more" which is horrible for screenreaders. E.g.: "Syncloud.com", "Get the UBOSbox", "More info on HomeDrive", etc. (They can all be the same format, but should include the actual product name.)
- The little lost heading "Currently available" is not needed. We only need to have the "Sold out or not available anymore" as a heading. And that should be bigger and centered. In the structure of the page, a h3 seems to make sense?
- Why is "Nextcloud Box" put in italics? :D

jospoortvliet:
Can't replace the images as I simply have no other images, but otherwise all changes done.

PR #1461, opened 2021-03-02
hide navigation bar for ios files app

blizzz:
it opens signup page, going around should not be possible


marinofaggiana:
@blizzz  where I can see the preview ?

blizzz:
> @blizzz where I can see the preview ?

This were having some issue, maybe some init-race-condition. 

obsoleted by #1462

PR #1464, opened 2021-03-11
The limit of 10 open documents has been removed from Collabora Online…

### LINKED ISSUE #1463: CODE docker limit is no more
On https://nextcloud.com/collaboraonline/ we can read:

`Issue: Collabora Online doesn't handle my 100 users.
This docker image is designed for home usage with a limited numbers of users and 10 open documents. If you need a more scalable solution, consider a support subscription for a reliable, business-ready online office experience.`

Since October, 2020 the limit of 10 documents has been removed.

timar:
… Development Edition (fixes #1463)


jospoortvliet:
I will re-add the item, to keep the "if you need a more scalable solution...' part, though ;-)

PR #1483, opened 2021-05-20
comment out ransomware sections

jospoortvliet:
Sadly, the very cool and useful ransomware protection app is currently undergoing a rewrite and not available until that is done. So for now, let's not point to it or talk about it, that can be seen as misleading.


LukasReschke (approved):
Thanks :)

PR #1489, opened 2021-06-02
Add RSS feed with alternatives

LukasReschke:
Ref https://github.com/nextcloud/security-advisories/discussions/25

@jospoortvliet WDYT? It's the old feed and a new entry:

> We have moved our new security advisories to GitHub Security Advisories. If you rely on a RSS feed, you can find alternatives discussed in https://github.com/nextcloud/security-advisories/discussions/25.

### DIFF
--- a/advisories/advisories.rss
+++ b/advisories/advisories.rss
@@ -1,0 +1,671 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+<channel>
+ <title>Nextcloud Security Advisories RSS Feed</title>
+ <link>https://nextcloud.com/security/advisories/</link>
+ <description>The Nextcloud security advisories as a RSS feed</description>
+ <ttl>1800</ttl>
+  <item>
+  <title>Security Advisories are moving to GitHub Security Advisories</title>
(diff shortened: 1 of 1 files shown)


jospoortvliet (approved):

m-ueberall:
Would it be possible to simply redirect [the old URI](https://help.nextcloud.com/t/security-advisories-rss-feed-broken-again/117625?u=m_ueberall) to the feed mentioned in nextcloud/security-advisories#25? This should minimize the required work on user side. (If I'm not mistaken, in this way, users should temporarily see some duplicates in the worst case.)

esbeeb:
I too would strongly prefer to receive Security Advisories over RSS, and it would be nice if no 3rd-party services were needed to make this happen (which can become a point of failure in and of itself, should the RSS feed stop working, as was the case for the feed mentioned in https://github.com/nextcloud/security-advisories/discussions/25)

Just to highlight that this is not an unreasonable thing to have, Gitlab, by comparison, makes a convenient RSS icon to subscribe to lists of issues.  An example (albeit, those are not security advisories) is seen [here](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues), which I happened across the other day.  Look in the upper right corner to see the RSS icon.

PR #1490, opened 2021-06-03
Add basic captcha

LukasReschke:


lib/captcha.php
```diff
@@ -0,0 +1,40 @@
+<?php
+
+if(session_status() !== PHP_SESSION_ACTIVE) {
+    session_start();
```
LukasReschke, line 4: This initiates a cookie but should qualify under "Essential Cookie" and does not require consent.

jospoortvliet (approved):

PR #1506, opened 2021-07-05
Fix header counter

marcoambrosini:
Signed-off-by: Marco Ambrosini <marcoambrosini@pm.me>

### DIFF
--- a/templates/header-top-navbar.php
+++ b/templates/header-top-navbar.php
@@ -1,53 +1,99 @@
 <?php $l = new L10N('header-navbar');
 require get_template_directory() . '/strings.php';
 ?>
-<!-- <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"> -->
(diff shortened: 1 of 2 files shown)


templates/header-top-navbar.php
```diff
@@ -1,53 +1,99 @@
 <?php $l = new L10N('header-navbar');
 require get_template_directory() . '/strings.php';
 ?>
-<!-- <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"> -->
+<!-- <meta http-equiv="refresh" content="60"> -->
```
juliushaertl, line 4: :fire:

juliushaertl (approved):
Seemed to work :+1:

templates/header-top-navbar.php
jospoortvliet: why did THAT happen? The refresh especially isn't nice ;-)

jospoortvliet (approved):
I can't speak on the rest but if Julius says it's good, it must be. Will test it at home and if ok, deploy.

PR #1508, opened 2021-07-13
add Carl

jospoortvliet:
Screenshot_20210713_105844


ognarb:
LGTM I will create another MR for the photo :)

jospoortvliet:
yeah, we usually take the pics in the office so they all look the same but if you have a similar looking pic, it's totally welcome until you visit us at some point and we can take one :D

PR #1542, opened 2021-09-05
fix: add rel=me to mastodon

SethFalco:
Minor pet peeve, but on Mastodon your site isn't verified because it doesn't have `rel="me"` in the HTML.

image

> If you put a link in your profile metadata, Mastodon checks if the linked page links back to your Mastodon profile. If so, you get a verification checkmark next to that link, since you are confirmed as the owner.
> 
> Behind the scenes, Mastodon checks for the `rel="me"` attribute on the link back. Likewise, Mastodon puts `rel="me"` on the links within profile metadata.
> — https://docs.joinmastodon.org/user/profile/

After this, it should display it as a verified link. That way users don't have to manually check the website to ensure this is the official Nextcloud account on Mastodon.

image


jospoortvliet (approved):

jospoortvliet:
works for me ;-)

PR #1548, opened 2021-09-22
Remove deprecated name attribute

Pytal:
Remove and replace the deprecated `name` attribute on anchor elements with `id`

As detailed in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-name

### DIFF
--- a/page-about.php
+++ b/page-about.php
@@ -138,7 +138,7 @@
             </ul>
         </p>
     </div>
-    <a name="history"></a>
+    <a id="history"></a>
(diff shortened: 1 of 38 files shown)


kesselb (approved):

splitt3r (approved):
Indentation looks strange sometimes. But i think that problem exists all over the code.

jospoortvliet (approved):
👍🏻 stuff still works 😄

PR #1553, opened 2021-10-01
Fix: update webchat link

mcnesium:
update webchat link from Riot to Element #610

### DIFF
--- a/page-translation.php
+++ b/page-translation.php
@@ -16,7 +16,7 @@
 <section class="content">
 <div class="container">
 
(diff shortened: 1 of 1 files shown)


jospoortvliet (approved):

jospoortvliet:
Only thing is - I'm not sure if that channel really is used anymore, we're active in Nextcloud Talk now... Hmmm. Still, let's update this, it's at least an improvement! Thanks! Meanwhile @rakekniven might know what the best way is to reach out to translators.

tcitworld:
There's people on Libera, but I don't see much going on Freenode.

jospoortvliet:
Then perhaps we should point to the community chat, @tcitworld  ??

tcitworld:
My opinion is that the help forums are probably the only viable thing, though I'm on most communication channels.

rakekniven:
Transifex is no place for discussions. 
Functionally, it is light years behind discourse.
The help forums are an active place. 
The channel at https://cloud.nextcloud.com/call/xhz52vhg is not actively used. People cannot create accounts themselves, or?

So I would vote for a link to the forums.

PR #1570, opened 2021-11-11
Update podcast-feed-m4a.rss

cartierbresson:
@ltsavar @jospoortvliet 🔍


ltsavar (commented):
the pub date is not correct <pubDate>Tue, 12 Nov 2021 9:00:00 GMT</pubDate>  .. should be thu not tue

jospoortvliet (approved):

PR #1578, opened 2021-11-30
fix link to Linux AppImage desktop client installer

mgallien:
Signed-off-by: Matthieu Gallien <matthieu.gallien@nextcloud.com>


mgallien:
please do not merge
I will rename the binaries because other stuff is broken with this name

PR #1601, opened 2022-02-01
jobs: Allow linking to specific opening

AlvaroBrey:
When sending a job opening to someone, I found it odd that there was no clear way to link a specific position. For example https://nextcloud.com/jobs/#androiddev just opens the page at the top and doesn't uncollapse the section.

With this patch, the JS uncollapses the section and scrolls to the header.

Note: I used the existing `jobs.js` file which appeared unused, but I'm not that familiar with this page's structure. I'll change that if needed.


jospoortvliet:
You are my hero, @AlvaroBrey

PR #1605, opened 2022-02-05
improve the website navigation and discoverability of install options

szaimen:
Four improvements:
1. improve order of options for "Get Nextcloud" menu
2. Change the description from server packages to Nextcloud server (I wouldn't know what server packages shall be)
3. Change the icon from a random icon to a download icon for this option which will make things more obvious and better to discover
4. in install-instructions menu, move appliances to be the first option since it should be the recommended way to install Nextcloud and not by using an archive or the webinstaller.

My basic idea about this PR was that a few years ago when I was new to Nextcloud and looked at the website to self-host my own server I didn't find the appliances because they are not easy to find. (Because of that I thought it was much too complicated for me and needed a few months until I had another look and found the appliances). This makes it hopefully much more easy to discover them.

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/install-instructions-server.php
+++ b/install-instructions-server.php
@@ -4,6 +4,14 @@
             <div class="row navigation-row">
                 <a href="#" class="close install-close">&times;</a>
                 <ul class="install-nav nav-tabs" role="tablist">
+                    <li>
+                        <a href="#tab-cloud" title="Provides automated updates" role="tab" data-toggle="tab">
+                            <div>
+                            <i class="fa-cloud"></i> <br />
+                            <?php echo $l->t('Appliances');?><br>
+                            </div>
+                        </a>
+                    </li>
                     <li id="li-tab-archive" class="active">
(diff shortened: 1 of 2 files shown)


szaimen:
@jospoortvliet please make sure test everything works as expected. I wasnt able to test my changes locally unfortunately...

jospoortvliet:
Hi,

So the clients-on-top, that was well thought-out: most users don't run a server but are users and need the clients; moreover, from the clients you can sign up (with simple signup). So we wanted to put that on top.

the other changes make sense. Could you undo the moving-clients-to-bottom in the menu? Then I'll merge :D

szaimen:
fine by me 👍

were you able to test my changes in install-instructions-server.php? 
(I haven't had the change to test if everything works as expected)

jospoortvliet:
Yeah it worked but I'll try one more time today and then merge it. Thanks!

PR #1626, opened 2022-03-27
Update for UBOSbox product refresh

jernst:
Also removed another duplicate listing, assuming that wasn't intentional.


jospoortvliet (approved):
makes sense indeed.

PR #1632, opened 2022-04-18
Replace PGP keyserver sks-keyservers.net with keys.openpgp.org

### LINKED ISSUE #1500: Broken instructions for PGP signature verification
The download page suggests getting the Nextcloud PGP key using this command:
```sh
gpg --keyserver ha.pool.sks-keyservers.net --recv-keys 28806A878AE423A28372792ED75899B9A724937A
```

The SKS-Keyserver network is going down. https://sks-keyservers.net/ has the following statement (if you ignore the certificate warning):
> This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.
> **Update 2021-06-21**: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

And indeed, `ha.pool.sks-keyservers.net` no longer resolves, breaking the signature verification instructions.
What is the preferred alternative now? `key.openpgp.org`? The key seems to be available there, but the identity is not verified.

Of course is is also still available on https://nextcloud.com/nextcloud.asc.
However, I personally would prefer some standardized solution that is directly supported by `gpg`. Aside from `keys.openpgp.org`, Web Key Directory (WKD) comes to mind. In a nutshell, WKD only requires hosting the key in a well-known location on `nextcloud.com` (see https://wiki.gnupg.org/WKD and https://wiki.gnupg.org/WKDHosting).

In summary I'd like to suggest doing at least one (but preferably both) of the following:
* verify the key's identity on `keys.openpgp.org`
* set up a WKD for the key

And, of course, the signature verification instructions should be changed accordingly.

tcitworld:
sks-keyservers.net is dead

Closes #1500


tflidd (approved):

tflidd:
It seems you can upload the key from Nextcloud side and verify by a mail address.

PR #1633, opened 2022-04-18
Bump CI docker images to PHP 7.4 and 8.0

tcitworld:
Obviously I don't know what's behind https://nextcloud.com but let's use something more civilized. :)


tcitworld:
Needs to update deps first :(

PR #1653, opened 2022-06-02
Add PiBox to devices page

pastudan:


pastudan:
Just updated the commit message so it passes DCO. This adds the PiBox product to the devices page.

For context, PiBox is our Raspberry Pi CM4 powered custom device, with support for 2 SATA SSD drives. We've tested NextCloud and it runs great on the PiBox, and is one of the most prominent apps in our app store.

Thanks for the review!

splitt3r:
It should be Nextcloud not NextCloud. At https://kubesail.com/template/erulabs/NextCloud you are using both.

Anyway looks like a cool project 👍🏻

